PR #48, opened 2015-01-21
Adds implementation of HTMLMediaElement, HTMLVideoElement, HTMLAudioElem...

jtakakura:
...ent and HTMLSourceElement


konsoletyper:
Please, write a small demo application that uses these interfaces

PR #61, opened 2015-02-16
adjust UnicodeSupport.mergePairs logic

wh0:
when we reach the end of one array, copy the rest of the other and return


konsoletyper:
Could you provide tests for this case?

PR #64, opened 2015-02-19
Fixed issue with null strings in $rt_str.

shannah:


konsoletyper:
Are you sure that $rt_str should check its argument for null? May be null check must be done externally?

shannah:
In my generated classes.js there are lots of instances like this:

~~~
function ccih_NetworkConnection1263_getResponseMessage4493($this) {
    return $rt_str($this.req1270.statusText);
}
~~~

This comes from the java code:

~~~
public String getResponseMessage() throws IOException {
        return req.getStatusText();
    }
~~~

Where getStatusText() could return null.

Issue #100, opened 2015-04-28
DataInputStream.readUTF Bug


shannah:
Malformed UTF byte error in DataInputStream.readUTF().  Patch forthcoming.

shannah:
I have create a pull request that fixes this issue.
https://github.com/konsoletyper/teavm/pull/101

PR #101, opened 2015-04-28
Fixed issue https://github.com/konsoletyper/teavm/issues/100

shannah:


konsoletyper:
Why did you move the following assignment:

```
out[s] = ch;
```

outside of `if` statement?

shannah:
Because this line:

~~~
} else if (((a = out[s]) & 0xe0) == 0xc0) {
~~~

requires that `out[s]` has already been set.

konsoletyper:
Hmm, I see. I don't like this code at all. May be I'll fix this better way a bit later.

Issue #112, opened 2015-05-13
TreeMap.remove() leaves tree in inconsistent state


shannah:
TreeMap.remove() sometimes deletes extra nodes.  Patch forthcoming.

shannah:
See pull request https://github.com/konsoletyper/teavm/pull/113

konsoletyper:
Can you provide a sequence of operations that causes this bug?

shannah:
~~~~
TreeMap<Double,Double> tm = new TreeMap<Double,Double>();
        for (double year = 1910; year <= 2010; year +=10) {
            tm.put(new Double(year), new Double(year));
        }
        for (double year = 1910; year <= 2010; year +=10) {
            System.out.println("Removing "+year);
            Double removed = tm.remove(new Double(year));
            System.out.println("Removed: "+removed);
            tm.put(new Double(year), new Double(year+1));
            System.out.println("Is 2000 still there?: "+tm.get(new Double(2000)));
        }
~~~~

Output:

~~~~
[EDT] 0:0:0,322 - Removing 1910.0
getData:1697 [EDT] 0:0:0,325 - Removed: 1910.0
getData:1697 [EDT] 0:0:0,328 - Is 2000 still there?: 2000.0
getData:1697 [EDT] 0:0:0,330 - Removing 1920.0
getData:1697 [EDT] 0:0:0,331 - Removed: 1920.0
getData:1697 [EDT] 0:0:0,335 - Is 2000 still there?: 2000.0
getData:1697 [EDT] 0:0:0,337 - Removing 1930.0
getData:1697 [EDT] 0:0:0,339 - Removed: 1930.0
getData:1697 [EDT] 0:0:0,341 - Is 2000 still there?: 2000.0
getData:1697 [EDT] 0:0:0,342 - Removing 1940.0
getData:1697 [EDT] 0:0:0,344 - Removed: 1940.0
getData:1697 [EDT] 0:0:0,346 - Is 2000 still there?: 2000.0
getData:1697 [EDT] 0:0:0,347 - Removing 1950.0
getData:1697 [EDT] 0:0:0,349 - Removed: 1950.0
getData:1697 [EDT] 0:0:0,352 - Is 2000 still there?: 2000.0
getData:1697 [EDT] 0:0:0,354 - Removing 1960.0
getData:1697 [EDT] 0:0:0,356 - Removed: 1960.0
getData:1697 [EDT] 0:0:0,357 - Is 2000 still there?: null
[...]

PR #113, opened 2015-05-13
Fix for https://github.com/konsoletyper/teavm/issues/112

shannah:
The problem was that rotateLeft() and rotateRight() returns the "new" root node for "this", so it it is called without its return value being inserted into the tree somewhere, it will disconnect the branch that was rotated from the rest of the tree.
This includes some extra housekeeping fixes to keep track of size.  
Note:  Balancing likely doesn't work correctly still as I opted to just make it "work" rather than figure out how to get balancing working properly.
(I.e. tree may get unbalanced)

### DIFF
--- a/teavm-classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
+++ b/teavm-classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
@@ -37,12 +37,12 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
                 if (right.factor() < 0) {
                     right = right.rotateRight();
                 }
-                return rotateLeft();
+                return this;
             } else if (factor == -2) {
                 if (left.factor() > 0) {
                     left = left.rotateLeft();
                 }
-                return rotateRight();
+                return this;
             } else {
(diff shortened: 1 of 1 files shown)


teavm-classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -37,12 +37,12 @@ public TreeNode(K key) {
                 if (right.factor() < 0) {
                     right = right.rotateRight();
                 }
-                return rotateLeft();
+                return this;
```
konsoletyper: Why? This kills balancing!

shannah: You can't call rotateLeft() or rotateRight() without assigning the result to the new root of the rotation otherwise it will result in the new root of the rotation to be removed.
This code by itself is not necessarily a problem.  It just place a condition, now, on the use of balance() that any call to balance() must assign the return value to the new root, and there were calls that did this (that have been removed). 

These changes do break balancing but it didn't look like there was a simple way to keep balance working while still making the tree function correctly without rewriting the whole thing.

PR #115, opened 2015-05-14
Fixed issue with size not being updated in TreeMap upon removal of so…

shannah:
…me nodes.

For issue https://github.com/konsoletyper/teavm/issues/112

### DIFF
--- a/teavm-classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
+++ b/teavm-classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
@@ -363,8 +363,8 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
             while (minDepth > 0) {
                 TreeNode<K, V> node = pathToMin[--minDepth];
                 node.left = right;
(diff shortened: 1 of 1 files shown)


teavm-classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -363,8 +363,8 @@ public boolean containsKey(Object key) {
             while (minDepth > 0) {
                 TreeNode<K, V> node = pathToMin[--minDepth];
                 node.left = right;
-                node.fix();
                 node = node.balance();
+                node.fix();
```
konsoletyper: This replacement is not necessary. balance() already fixes heights.

shannah: If the node is already balanced, then balance() won't touch the height.

konsoletyper:
Yes, but heigts are invalid before balance(). So you should not have removed it.

shannah:
I see your point.  This particular change may not have been necessary.  I see your change, and I'm in the process of testing it.

PR #122, opened 2015-06-02
Implemented Thread.join()

shannah:


konsoletyper:
Could you please rewrite this in terms of Async methods?

shannah:
I'm not sure exactly how without duplicating the implementation of Object.wait().  Instead of using a lock object and notifyAll(), I could keep an ArrayList of listeners that I loop through after run completes, but this wouldn't improve performance and it would involve quite a bit more code, and I'll need to test to make sure I'm switching back to the proper thread in each listener for the async callback to complete -- all of which is handled by wait() with the current implementation.

Maybe I don't exactly understand what you have in mind.

shannah:
Thanks for merging it.  If you can give me a little more direction, I'm happy to change the implementation still.

konsoletyper:
I think you were right. While `Thread.join()` without arguments is pretty simple, `Thread.join(long)` might be quite painful. The only disadvantage of your implementation is that it can cause additional 'context switch' at the end of a thread in rare cases. Also, it gives some overhead, but I hope it won't be very significant.

PR #139, opened 2015-07-21
Fix for Character.getNumericValue.

mpoindexter:
UnicodeHelper.encodeIntByte/decodeIntByte produced garbage character metadata
since 'z' + [some high byte value] produced a character greater than 255, which
did not properly survive the encoding/decoding process.

Instead, use Base46.encode/decode which uses only printable characters in the string.

### DIFF
--- a/teavm-classlib/src/main/java/org/teavm/classlib/impl/unicode/UnicodeHelper.java
+++ b/teavm-classlib/src/main/java/org/teavm/classlib/impl/unicode/UnicodeHelper.java
@@ -17,6 +17,9 @@ package org.teavm.classlib.impl.unicode;
 
 import java.util.Arrays;
 
+import org.teavm.classlib.impl.Base46;
+import org.teavm.classlib.impl.CharFlow;
+
 /**
  *
  * @author Alexey Andreev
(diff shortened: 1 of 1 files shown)


mpoindexter:
I looked into this a bit further, and it seems like the original code works, but only under some circumstances:  namely, if the hosting page has an encoding of UTF-8.  The auto-generated html page created by the maven plugin does not.  The junit test page does.  If the hosting page has an encoding other than UTF-8, Character.getNumericValue returns garbage results because the string is not interpreted correctly.  

Maybe it's just a requirement that the hosting page be UTF-8 (or setting the charset at utf-8 via a Content-Type header on the JS resource itself would probably work as well, but I did not test it), if so this should probably be fixed in the auto-generated sample page, or documented somewhere.  

It seems to me that it might be nice for the generated JS to not be sensitive to the encoding the page is served in though, in which case this patch will help.  I would not expect something like Integer.parseInt("1234") to fail just because a JS file was not served with a specific charset in the MIME type.  There are still test cases that fail if junit.html is not UTF-8, but this change reduces the failures from > 300 to 20.  It seems like several of the failures are just assertion failures due to the hardcoded test value being incorrectly interpreted due to the encoding, but I haven't looked into it fully.

konsoletyper:
I agree with you that this method can be rewritten to not use UTF-8. Another reason for this: ASCII characters with code less than 128 require only one byte in UTF-8, so it can improve compression rate. I am disagree only in that you didn't go futher and use better compression algorithm. The idea is the following. We need to compress (character, numeric value) pairs. These pairs consist of consequent characters and consequent numeric values, and there are gaps for ranges of characters which don't have any numeric value. So we can encode two types of objects:

1. gap, as a single non-negative number.
2. increasing sequence of pairs, with the given length and starting digit, both non-negativ

Please, notice, that Base46 produces shorter sequence with `encodeUnsigned`, since it haven't to store sign. I can propose the following format:

1. Odd number *n* means gap, *n* / 2 is the length of the gap.
2. Even number *n* means sequence, *n* / 2 is the length of the sequence, and the following number is the starting digits.

I wanted to implement it by myself, but I have many other things in my roadmap, so I have no time for this issue. If you could implement this algorithm or invent any similar idea, which produces both small sequence and decompression code, your efforts would be greatly appreciated.

As for UTF-8, there are string literals, that are still encoded in UTF-8. I discussed this and even proposed a workaround (i.e. generating \uNNNN sequences for all chars > 127), but still haven't implemented this. I'm not sure if it really necessary, since nowadays there's no problems to set up a proper character encoding.

mpoindexter:
I can do that.  

It's worth mentioning that with my change as is it's already roughly half of what it was (2421 bytes vs 4158 bytes when encoded as UTF-8) since Base46 uses a variable length encoding already and lots of the numbers are short.  Switching to unsigned reduces it even further to 2216 bytes.

I guess I wonder if it's worth doing heroics to get even smaller...it seems like it's going to be gzip'd on the wire anyway, and the gzip compressor can do pretty much the same RLE without having to do silly things like stay in the ASCII character range.  It might actually be better to leave it relatively uncompressed:  if the file is served gzip'd by the web server, it means that decompression will be done in fast native code by the web browser, and any code that is not run in JS will improve the startup time.

Anyway, I will make the change and we will see if it's worth it.

konsoletyper:
Ok, I'll think about it. As for generated JS size, there are people who think that JS file itself should be as small as possible. Rather than convincing them, I try to keep TeaVM generating small JS files. But in this case there's no need to reduce the generated JS by only, say 1kb.

PR #140, opened 2015-07-21
Fix the order comparison is applied in for binarySearch to match the …

mpoindexter:
…JDK.

This is necessary since accoding to the method signature of binarySearch
only the elements of the list must implement Comparable<T>, not the key
itself.

### DIFF
--- a/teavm-classlib/src/main/java/org/teavm/classlib/java/util/TCollections.java
+++ b/teavm-classlib/src/main/java/org/teavm/classlib/java/util/TCollections.java
@@ -259,10 +259,10 @@ public class TCollections extends TObject {
         while (true) {
             int i = (l + u) / 2;
             T e = list.get(i);
-            int cmp = -1 * c.compare(e, key);
+            int cmp = c.compare(e, key);
             if (cmp == 0) {
                 return i;
-            } else if (cmp < 0) {
+            } else if (cmp > 0) {
                 u = i - 1;
                 if (u < l) {
                     return -i - 1;
--- a/teavm-classlib/src/main/java/org/teavm/classlib/java/util/TCollections.java
+++ b/teavm-classlib/src/main/java/org/teavm/classlib/java/util/TCollections.java
@@ -259,7 +259,7 @@ public class TCollections extends TObject {
         while (true) {
             int i = (l + u) / 2;
             T e = list.get(i);
-            int cmp = c.compare(key, e);
+            int cmp = -1 * c.compare(e, key);
             if (cmp == 0) {
                 return i;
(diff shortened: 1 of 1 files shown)


konsoletyper:
My proposal is the following: don't change `binarySearch` itself, rather fix `naturalOrder`. For example, you could replace `o1 != null` with `o1 instanceof TComparable<?>`.

mpoindexter:
Don't think changing the comparator is right.  In Oracle JDK the element from the list is always the left hand side of the comparison.  I think it should be the same in TeaVM for compatibility.

I encountered a problem in code that looked like this:

     Foo implements Comparable<Integer>
     List<Foo> lst = new ArrayList<Foo>();
     int indexOfFooMatching1234 = Collections.binarySearch(lst, 1234);

Obviously lst.get(0).compareTo(1234) would return a different result than 1234.compareTo(lst.get(0)).  We can leave debating whether this is a good idea aside...it's not my code and I wouldn't write it like that, but it is a case that works in the Oracle JVM (and seems to be implied by the method signature that it should work).  I think it should be the same if a comparator is passed (the element from the list is the lhs of the compare), so changing it in the comparator is not the best.

konsoletyper:
Oh, depending on such shitty code is like hell. I'm not sure TeaVM needs this, since is can't be 100% compatible with Oracle JVM (there are limitations like lack of reflection, no array bound checks, no NPE on trying to call method on `null`). Anyway, multiplying by -1 is not the best solution. You could just use unary minus, or even better, invert comparison operator.

BTW, what is the project you are trying to compile with TeaVM?

mpoindexter:
Yeah, I was a bit surprised when I saw the code, but looking at the signature of Collections.binarySearch it seems to me that it's intended to work.  Here's the signature:

     public static <T> int binarySearch(List<? extends Comparable<? super T>> list, T key)

This signature implies to me that the elements in list must be comparable to T, but that T is not necessarily comparable to the elements in the list (maybe I am reading it wrong, but I would expect that if T had to be comparable to the elements in the list the sig would be something like:

     public static <T> int binarySearch(List<? extends Comparable<? super T>> list,  Comparable<T> key)

As for what I am trying to compile, I'm actually trying to do something a bit crazy:  I am doing some experiments with Media Source Extensions in the browser, so I've got a fork of ExoPlayer (https://github.com/google/ExoPlayer) with all the Android specific stuff stripped out that I am using to feed the browser's media pipeline.  It's surprisingly actually working, I was pleasantly surprised by how much of Java TeaVM manages to handle!

mpoindexter:
Inverted the comparison instead of multiplying by -1

PR #141, opened 2015-07-21
Implement System.console().  Since this method is defined to return null

mpoindexter:
if no console is available, return null.


konsoletyper:
Please, add copyright headers

mpoindexter:
Added copyright headers

PR #142, opened 2015-07-21
Various fixes to ByteBuffer

mpoindexter:
-Fix exception when put()'ing a zero byte array
-Fix getInt() methods returning a short value
-Fix relative get() using the index instead of the supplied position
-Add long put/get


konsoletyper:
Please, add tests. You should run them both from JVM and TeaVM. For the latter simply build `teavm-tests` and open `target/javascript-tests/junit.html` in the browser. Or use command line tool to build one tests: `org.teavm.cli.TeaVMTestRunner -d <target_dir> <test_class>`, and don't forget to include all necessary jars into classpath.

mpoindexter:
Added tests.  They pass on TeaVM and JVM.

PR #144, opened 2015-07-21
Add MediaSourceExtension elements

mpoindexter:


teavm-dom/src/main/java/org/teavm/dom/media/MediaSource.java
```diff
@@ -0,0 +1,24 @@
+/*
+ *  Copyright 2014 Alexey Andreev.
```
konsoletyper: It's 2015!

mpoindexter:
Fixed

PR #145, opened 2015-07-21
Implement Class.getSimpleName

mpoindexter:


konsoletyper:
Where are tests? What about arrays? What about primitives?

mpoindexter:
Added tests and fixed arrays.  Primitives were already correct.

PR #149, opened 2015-08-29
Interrupted exception in tests

jtulach:
Here is raw implementation of *InterruptedException* behavior as expected by the HTML/Java TCK tests. Seems to fix at least one test.


konsoletyper:
See my comment for pull request 150

jtulach:
I see. It is easy for me to modify existing releases. However the patch should be mergeble to master. This is the actual patch: https://github.com/konsoletyper/teavm/compare/0.3.1...jtulach:InterruptedExceptionInTests

PR #150, opened 2015-08-30
Put people uses right method

jtulach:
I am able to fix the test by doing the substitutions during a request time and by returning String rather than JSON when text/plain mimetype is requested.


konsoletyper:
I can't merge this pull request since it includes branch 0.3.x, which will never be merged into master. I usually use diff files or cherry picking to transfer changes from release branches to master and vice versa.

jtulach:
I see. It is easy for me to modify existing releases. However the patch should be mergeble to master. This is the actual patch: https://github.com/konsoletyper/teavm/compare/0.3.1...jtulach:InterruptedExceptionInTests

jtulach:
Updated the patch to be based on origin/master - hopefully it will be easily mergeable now.

PR #162, opened 2015-10-25
add jso File interface and update input element

leobm:


leobm:
Hi,

there is an error in the build test.

```
Driver info: org.openqa.selenium.remote.RemoteWebDriver

Exception in thread "Thread-7" org.openqa.selenium.UnsupportedCommandException: Sauce Labs Authentication Error.

You used username 'None' and access key 'None' to authenticate, which are not valid Sauce Labs credentials.
```

konsoletyper:
I see. All pull requests fail tests. It seems that passwords I specified for Sauce Labs selenium works only from my travis account. I'll try to reconfigure it later

PR #171, opened 2015-12-24
Initial attempt to support resources in TeaVM

jtulach:
Here is an initial sketch of the support for resources in TeaVM as I envision it.

There is a Renderer that find needed resources and dumps their content into the generated JavaScript file. The TClass.getResourceAsStream is modified to find the resources and use them.

TODO:
* window.teaVMResources is certainly not the right object to store resources at - the object should be private to the generated JavaScript, not visible from outside.
* Implement other resource reading methods
* How to configure ResourcesInterceptor to know what resources to encode?
* How can ResourcesInterceptor scan the classpath for list of available resources?

Thanks in advance, @konsoletyper, for any advices.


samples/benchmark/pom.xml
```diff
@@ -17,6 +17,10 @@
```
jtulach: Ignore the following files. I guess I messed up when using Git. I'll fix it during next iteration. I am sorry.

konsoletyper:
I merged this branch to release-0.4.x. Will create next minor release of TeaVM quite soon. Everything OK, except for I'd prefer to make list of copied resources a little more customizable. I'll do it myself.

PR #173, opened 2015-12-27
Declare resources in pom.xml

jtulach:
It makes more sense for the applications using TeaVM (like https://github.com/controlsjs/controls4j) to specify the resources to include in the generated script in pom.xml. It is easier for them and their end-users to control pom.xml than write their own Java written supplier.

### DIFF
--- a/html4j/src/main/java/org/teavm/html4j/HTML4jResourceSupplier.java
+++ b/html4j/src/main/java/org/teavm/html4j/HTML4jResourceSupplier.java
@@ -15,8 +15,7 @@
  */
 package org.teavm.html4j;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.io.File;
 import org.teavm.classlib.ResourceSupplier;
 import org.teavm.model.ListableClassReaderSource;
 
@@ -27,17 +26,10 @@ import org.teavm.model.ListableClassReaderSource;
 public class HTML4jResourceSupplier implements ResourceSupplier {
     @Override
     public String[] supplyResources(ClassLoader classLoader, ListableClassReaderSource classSource) {
-        Set<String> resources = new HashSet<>();
-        for (String className : classSource.getClassNames()) {
-            final int lastDot = className.lastIndexOf('.');
-            if (lastDot == -1) {
-                continue;
-            }
-            String packageName = className.substring(0, lastDot);
-            String resourceName = packageName.replace('.', '/') + "/" + "jvm.txt";
-            resources.add(resourceName);
+        String resources = System.getProperty("resourcesSupplier");
+        if (resources == null) {
+            return null;
         }
-
-        return resources.toArray(new String[0]);
+        return resources.split(File.pathSeparator);
     }
 }
--- a/html4j/pom.xml
+++ b/html4j/pom.xml
@@ -81,6 +81,7 @@
               <targetDirectory>${project.build.directory}/javascript-test</targetDirectory>
               <debugInformationGenerated>true</debugInformationGenerated>
               <sourceMapsGenerated>true</sourceMapsGenerated>
+              <resources>**/*.txt</resources>
             </configuration>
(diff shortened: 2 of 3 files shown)


html4j/src/main/java/org/teavm/html4j/HTML4jResourceSupplier.java
```diff
@@ -27,17 +26,10 @@
...
-        Set<String> resources = new HashSet<>();
-        for (String className : classSource.getClassNames()) {
-            final int lastDot = className.lastIndexOf('.');
-            if (lastDot == -1) {
-                continue;
-            }
-            String packageName = className.substring(0, lastDot);
-            String resourceName = packageName.replace('.', '/') + "/" + "jvm.txt";
-            resources.add(resourceName);
+        String resources = System.getProperty("resourcesSupplier");
```
jtulach: System property isn't the best way to propagate information between Mojo and the supplier, but it was the easiest. There is a quite a lot of TeaVM layers in between and I am not sure what is the best way to cross through them.

html4j/src/main/java/org/teavm/html4j/HTML4jResourceSupplier.java
```diff
@@ -27,17 +26,10 @@
...
     @Override
     public String[] supplyResources(ClassLoader classLoader, ListableClassReaderSource classSource) {
-        Set<String> resources = new HashSet<>();
-        for (String className : classSource.getClassNames()) {
-            final int lastDot = className.lastIndexOf('.');
-            if (lastDot == -1) {
-                continue;
-            }
-            String packageName = className.substring(0, lastDot);
-            String resourceName = packageName.replace('.', '/') + "/" + "jvm.txt";
```
jtulach: The jvm.txt was just an example, it doesn't have any special meaning in the HTML/Java API neither in the Controls4j applications.

konsoletyper:
I disagree. User should not specify resource suppliers at all. Instead, frameworks like controls.js could code with supplier class together with corresponding resource. Alternative approach is to provide a separate jar file named "compatibility layer with TeaVM" that contains this class.

It's too late to change, since I've just released TeaVM with your previous pull request recently.

jtulach:
Can you comment on the proper way to propagate information from the pom.xml to the ResourceSupplier which wouldn't involve System property?

konsoletyper:
There's no such way

jtulach:
That is a limitation. suppliers without knowing anything about the user configuration are a bit limited. They don't even know where to scan for available resources.

What would be your preferred way to list all resources on a classpath and make them available to ResourceSuppliers?

konsoletyper:
> What would be your preferred way to list all resources on a classpath and make them available to ResourceSuppliers

Just take a look at ResourceSupplier interface. The only thing that I really forgot is to provide properties from pom.xml configuration.

konsoletyper:
Check my latest commit concerning resources before I publish release 0.4.3: https://github.com/konsoletyper/teavm/commit/539a12227f79abdb4ce22aac2dbc647f1c09f083

Instead of parameters I pass a single object to ResourceSupplier, this allows to extend this API without breaking binary compatibility.

jtulach:
Hello Alexey,
My cousin reported to me that his controls4j will be running on 0.4.3
thanks to your base64 fix.

Dne 28. 12. 2015 19:03 napsal uživatel "Alexey Andreev" <
notifications@github.com>:
>
> Check my latest commit concerning resources before I publish release
0.4.3: 539a122
>
> Instead of parameters I pass a single object to ResourceSupplier, this
allows to extend this API without breaking binary compatibility.

#1 - My cousin already had his implementation so that one would be broken.

#2 - the "single object" should be non-subclassable to be 100% sure:
http://wiki.apidesign.org/wiki/RequestResponse

But that's just nitpicking. Thanks for helping Jirka.
-jt

PR #179, opened 2016-01-28
Let TeaVMHost know about the build parameters 

bertung:
I made two useful additions to the TeaVMHost which would help developers to create new files to the target build directory according to the annotations or other class properties set. This is very useful especially for distributed systems that would need custom bootstrapper or initiator.

For example we have an annotation called @Remote, If annotation exists on the Application class we create extra html and javascript files that would run remotely (this is a special case for our platform). In short, it is very useful for complex use cases. I also reformatted my previous class lib additions. You are free to reject them all though :).


konsoletyper:
Why are you creating a pull instead of asking my opinion and discussing?  I think that TeaVM should not provide some information about abolute location of a generated file. I can't understand why generate HTML files from TeaVM at all.

Why do you need these classes in JDK? Can't you avoid them? I can't accept some of the classes, like TFile, since I follow the goal: once TeaVM supports something from JDK, is must provide as complete support as possible. Imagine a user who wants TeaVM for his purpose and is glad to see java.io.File among supported classes. But then he is disappointed since File is just a stub.

You can ether propose fully-featured implementations, but some discussion required. Where are you going to store files? How do you solve the problem of async API? What browsers are supported by your code? What impact does you code has with threads? Another option for you is to either replace these classes in your code base by ones supported by TeaVM. The third option is to create plugin for TeaVM, that introduces ClassHolderTransformer that automatically performs such replacement, without changing the initial code base. The fourth option is to maintain your own fork of TeaVM and your own binary builds in private Maven repository, just like CodeNameOne did.

Also, please note that whenever I approve pull request, it won't mean that I will release a new version with proposed changes. Now I put all my efforts into TeaVM Flavour project and I am not going to make major releases of TeaVM soon.

bertung:
You are right, my bad. I rarely create pull requests. Generating files through annotations are for code accessibility purposes for the developer that is something we need and thought would be useful for others as well. Class files are mostly interfaces, and meant to compile existing applications without errors nothing more. They are simple classes, dont think they would create any runtime issues. Anyways, I will work on my fork but build fails. I opened an issue for it.

PR #188, opened 2016-03-07
Implemented Gradle plugin.

czyzby:
Gradle plugin based on [Kotlin version](https://github.com/edibleday/teavm-gradle-plugin). Translated to Java on request for code consistency. Has more settings (including TeaVM version, which is fixed in former plugin) and fixes minor bugs (output folders are taken directly from `Project` settings rather than hardcoded to `build/classes/main` and `build/resources/main`). Includes an example project.

### DIFF
--- a/tools/gradle/.gitignore
+++ b/tools/gradle/.gitignore
@@ -1,0 +1,6 @@
+bin/
+.classpath
+.project
+.settings/
+.gradle/
+build/
--- a/tools/gradle/README.md
+++ b/tools/gradle/README.md
@@ -1,0 +1,33 @@
+# Gradle plugin for TeaVM
+
+Somewhat based on [the Kotlin version](https://github.com/edibleday/teavm-gradle-plugin), refactored and translated to Java for TeaVM code consistency.
+
+## Using the plugin
+
+Add build script dependency:
+```
+buildscript {
+    repositories {
+        mavenLocal()
+        mavenCentral()
+    }
+    dependencies {
+        classpath "org.teavm:teavm-gradle-plugin:0.4.3"
+    }
+}
+```
+
+Apply plugin:
+```
+apply plugin: "teavm"
+```
+
+Use `compileTeaVM` task to compile your application.
+
+## Working with the source
+
+Run `gradle eclipse` or `gradle idea` to generate IDE project.
+
+### Publishing
+
+Use `gradle build install` to publish the plugin into local Maven repository. If you want to publish to Maven Central, fill `gradle.properties` file with signing and logging data. (Be careful not to commit the changes! Hint: you can keep a separate `gradle.properties` file with your Maven Central data in your [Gradle home folder](http://mrhaki.blogspot.com/2015/10/gradle-goodness-setting-global.html).) Use `gradle uploadArchives` to publish the artifact. Changing the `isSnapshot` property to `-SNAPSHOT` or '' (empty string) allows to choose whether a snapshot or final version should be published.
--- a/tools/gradle/build.gradle
+++ b/tools/gradle/build.gradle
@@ -3,6 +3,7 @@ apply plugin: 'eclipse'
 apply plugin: 'idea'
 apply plugin: 'maven'
 apply plugin: 'signing'
+apply plugin: 'checkstyle'
 
 sourceCompatibility = 1.6
 
--- a/tools/gradle/build.gradle
+++ b/tools/gradle/build.gradle
@@ -1,0 +1,111 @@
+apply plugin: 'java-gradle-plugin'
+apply plugin: 'eclipse'
+apply plugin: 'idea'
+apply plugin: 'maven'
+apply plugin: 'signing'
+
+sourceCompatibility = 1.6
+
+ext {
+    libVersion = '0.4.3'
+    isSnapshot = '-SNAPSHOT'
+}
+
+group = projectGroupId
+archivesBaseName = "teavm-gradle-plugin"
+version = "$libVersion$isSnapshot"
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+    jcenter()
+}
+
+configurations {
+    deployerJars
+}
+
+dependencies {
+    compile gradleApi()
+    compile "org.teavm:teavm-core:$libVersion"
+    compile "org.teavm:teavm-tooling:$libVersion"
+    deployerJars "org.apache.maven.wagon:wagon-ssh:2.2"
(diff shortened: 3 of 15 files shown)


konsoletyper:
1. Is it possible to build Gradle plugin with Maven? The entire TeaVM project is built with Maven by simply running "mvn install" command. Gradle plugin needs separate build command, which looks strange to me.
2. Most of TeaVM modules (except classlib, which I mostly copy from Harmony) use checkstyle. Gradle plugin should be checked with checkstyle as well. See `core` module for example.
3. Why don't parameter defauts conform with Maven plugin? For example, `targetFile` default value must be `classes.js`.
4. `cacheDirectory` does not work in non-incremental mode. I could not find incremental mode support in Gradle plugin.
5. `properties` are not supported as well. Gradle plugin must support it. Without `properties` you can't customize locales, for example.
6. Why does `gradle.properties` file contains properties like `signing.keyId`? Credentials may not be in source code repositories. They must be stored privately.
7. What about Eclipse integration? When you import Maven project in Eclipse, it automatically configures Eclipse TeaVM plugin (see [TeaVM m2e plugin](https://github.com/konsoletyper/teavm/tree/master/tools/eclipse/m2e-plugin)). The same feature should be available with Gradle.

czyzby:
1. Using Maven for Gradle plugin might be tricky, I don't know the exact dependencies for Gradle API. If you're not sure how to convert it to Maven yourself, we might need to do some research on this.
2. There is a `checkstyle` plugin to Gradle, I can look into this until Maven is used.
3. I based it on Kotlin plugin, which used `app.js`, if I'm not mistaken. I'll correct the default values.
5. Which properties?
6. They are commented out. Gradle tasks will not finish unless these properties are present in a `.properties` file. This is simply a utility for contributors which do not have these globally set - uncommenting those will all to use Gradle tasks without any errors.
7. IDEs usually support running Gradle tasks "natively" and can detect these, but it's up to the Gradle support in IDE, really. I don't think another Eclipse plugin will be required.

konsoletyper:
`4.` No, TeaVM incremental build does not integrate with external tools like Maven and Gradle. All you need is to pass correct parameters to `TeaVMTool`. Please, learn teavm-maven-plugin source code.
`5.` [Here](https://github.com/konsoletyper/teavm/wiki/Building-JavaScript-with-Maven-and-TeaVM), note `properties` build parameter.
`7.` There is [TeaVM plugin for Eclipse](https://github.com/konsoletyper/teavm/wiki/Eclipse-tutorial). Please, try to install and learn it. Running TeaVM from Eclipse != running it from Gradle or Maven. For example, it puts markers into Java editor when TeaVM reports an error. TeaVM plugin for m2eclipse does not run TeaVM, instead it configures TeaVM plugin for eclipse. I suppose, Gradle must do something similar.

czyzby:
I see, I shouldn't have based it on Kotlin plugin so much. Thanks for the link about Maven plugin, I'll try implementing all listed options (including class and method aliases).

As for the integration with Eclipse plugin, I'm afraid you'll have to look into it yourself. I'm not comfortable with Eclipse API at all, neither do I plan to use this IDE with TeaVM.

czyzby:
Now all properties from Maven task (excluding `classFiles`) are supported by the Gradle plugin. Default values match the documentation from the page you posted. Checkstyle plugin added, code should pass the validation. Compiling works as expected, but running the compiler in incremental mode causes unexpected internal error:
```
    Unable to compile TeaVM application.
    java.lang.IndexOutOfBoundsException: Index: 9, Size: 0
        at java.util.ArrayList.rangeCheck(ArrayList.java:653)
        at java.util.ArrayList.get(ArrayList.java:429)
        at org.teavm.cache.FileSymbolTable.at(FileSymbolTable.java:81)
        at org.teavm.cache.DiskCachedClassHolderSource.readClass(DiskCachedClassHolderSource.java:117)
        at org.teavm.cache.DiskCachedClassHolderSource.get(DiskCachedClassHolderSource.java:58)
        at org.teavm.cache.DiskCachedClassHolderSource.get(DiskCachedClassHolderSource.java:27)
        at org.teavm.dependency.DependencyClassSource.findClass(DependencyClassSource.java:79)
        at org.teavm.dependency.DependencyClassSource.findAndTransformClass(DependencyClassSource.java:68)
        at org.teavm.dependency.DependencyClassSource.lambda$get$24(DependencyClassSource.java:51)
        at java.util.HashMap.computeIfAbsent(HashMap.java:1118)
        at org.teavm.dependency.DependencyClassSource.get(DependencyClassSource.java:51)
        at org.teavm.dependency.DependencyClassSource.get(DependencyClassSource.java:37)
        at org.teavm.model.ClassReaderSource$2.add(ClassReaderSource.java:88)
        at org.teavm.model.ClassReaderSource$2.<init>(ClassReaderSource.java:61)
        at org.teavm.model.ClassReaderSource.lambda$getAncestors$17(ClassReaderSource.java:56)
        at java.lang.Iterable.spliterator(Iterable.java:101)
        at org.teavm.model.ClassReaderSource.getAncestors(ClassReaderSource.java:94)
        at org.teavm.model.ClassReaderSource.resolve(ClassReaderSource.java:98)
        at org.teavm.model.ClassHolderSource.resolveMutable(ClassHolderSource.java:37)
        at org.teavm.dependency.DependencyChecker.lambda$new$0(DependencyChecker.java:93)
        at org.teavm.common.CachedMapper.map(CachedMapper.java:52)
        at org.teavm.dependency.DependencyChecker.linkMethod(DependencyChecker.java:318)
        at org.teavm.vm.TeaVM.entryPoint(TeaVM.java:252)
        at org.teavm.tooling.TeaVMTool.generate(TeaVMTool.java:312)
        at org.teavm.gradle.task.CompileTask.compileTeaVM(CompileTask.java:70)
        at org.teavm.gradle.task.CompileTask.compile(CompileTask.java:55) 
        // Rest omitted, Gradle related.
```
Not sure if that's an error on my part or a bug in `0.4.3` - any ideas?

czyzby:
[Building the plugin with Maven will not be as straightforward](https://discuss.gradle.org/t/build-gradle-plugins-with-maven/7464/12). However, there is a [Gradle plugin for Maven](https://github.com/LendingClub/gradle-maven-plugin), which allows to invoke Gradle tasks. You would just have to bind `mvn install` to `gradle uploadArchives`/`gradle build install` (the first uploads to Maven snapshots/central depending on version, the second to local repo), and possibly any other required tasks - like `gradle checkstyle`.

PR #194, opened 2016-06-24
provide missing info() methed on Logger

stevep99:
Added missing Logger.info(), similar to existing warning() and config() methods.
The private internal method names renamed to avoid clash.


Ihromant:
This is outdated PR and in can be closed.

PR #215, opened 2016-08-18
Use max() instead of min() when enlarging the array.

sba1:
Adding n elements subsequently has now an O(n) running time while it was
O(n^2) before.


sba1:
> Merged #215 [1].

Thanks for merging. This may also be something for 0.4.x (if it is still 
maintained).

konsoletyper:
No, it's not maintained anymore

PR #218, opened 2016-08-23
Added missing classes/methods from Java SE Runtime

skapral:


classlib/src/main/java/org/teavm/classlib/java/lang/TStackOverflowError.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2016 skapral.
```
konsoletyper: I believe you should specify your real name here.

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
```diff
@@ -43,7 +46,7 @@
     V put(K key, V value);
 
     V remove(Object key);
-
+
```
konsoletyper: Remove redundant whitespaces

konsoletyper:
Please, add tests for Map methods you added

konsoletyper:
Ok, I merged your PR, but FYI, you should have created a separate file to write tests (since there used to be one-to-one relationship between JDK classes and tests). Also, don't name test method like `test_X`. There are tests that I got from Harmony, and it used ancient JUnit 3, which relies on such naming convention. With JUnit 4 you have `@Test` annotation, so you don't need to prefix your methods' names with `test`.

PR #238, opened 2017-02-08
Fix asynchronous instantiation via reflection. 

davmac314:
The generated code resumes from a thread suspension event, but won't appropriately re-suspend if the
constructor suspends; this is a fix.

### DIFF
--- a/tests/src/test/java/org/teavm/classlib/java/lang/TestObjectAsync.java
+++ b/tests/src/test/java/org/teavm/classlib/java/lang/TestObjectAsync.java
@@ -22,8 +22,8 @@ public class TestObjectAsync extends Object implements Runnable {
 
     public TestObjectAsync() {
         try {
-            Thread.sleep(100);
-            Thread.sleep(100);
+            Thread.sleep(1);
+            Thread.sleep(1);
             counter = 1;
         }
         catch (InterruptedException ie) {
--- a/tests/src/test/java/org/teavm/classlib/java/lang/TestObjectAsync.java
+++ b/tests/src/test/java/org/teavm/classlib/java/lang/TestObjectAsync.java
@@ -1,0 +1,42 @@
+/*
+ *  Copyright 2017 Alexey Andreev.
+ *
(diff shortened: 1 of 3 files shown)


davmac314:
@konsoletyper I have created a test, however the test fails with and without the previous commit due to an unrelated problem. To see the test pass (with the previous commit), add "Object.class.newInstance()" at the beginning of the new test (instanceCreatedThroughReflectionAsync()) - this works around the other problem, which is caused by the Platform class newInstance method being swapped out for a different method (in Platform.preparenewInstance()), which invalidates the thread suspend/resume stack.

tests/src/test/java/org/teavm/classlib/java/lang/TestObjectAsync.java
```diff
@@ -0,0 +1,42 @@
...
+package org.teavm.classlib.java.lang;
+
+public class TestObjectAsync extends Object implements Runnable {
+    private int counter;
+    
+    public static Runnable r;
+
+    public TestObjectAsync() {
+        try {
+            Thread.sleep(100);
```
konsoletyper: Is it really necessary to wait 100 milliseconds? Why not `Thread.sleep(1)`. Why not `Thread.yield()`?

davmac314: > You don't need this line:

Ok; I was just mirroring the previous test. I have now changed it to use a class literal as you suggested.

> Is it really necessary to wait 100 milliseconds? Why not Thread.sleep(1)? Why not Thread.yield()?

True, Thread.sleep(1) is sufficient (now committed). Thread.yield() however does not trigger failure. The implementation of Thread.yield() appears to do nothing unless it is called at least 30 times:

```
   public static void yield() {
        TThread currentThread = currentThread();
        if (++currentThread.yieldCount < 30) {
            return;
        }
```

I suspect that many of the other tests, which do use Thread.yield(), would be more thorough if they instead used Thread.sleep(1).

PR #240, opened 2017-02-09
Fix first-time instantiation via reflection.

davmac314:
In 7fc035f, a general fix to instantiation via reflection was made and a
test case added. The test case fails due to a separate issue, solved in
this commit, which affects only the first instantiation via reflection.

### DIFF
--- a/platform/src/main/java/org/teavm/platform/Platform.java
+++ b/platform/src/main/java/org/teavm/platform/Platform.java
@@ -33,6 +33,8 @@ public final class Platform {
     private Platform() {
     }
 
+    private static boolean newInstancePrepared;
+
     @InjectedBy(PlatformGenerator.class)
     public static native PlatformObject getPlatformObject(Object obj);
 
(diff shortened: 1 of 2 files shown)


davmac314:
This is one possible solution, though it removes a minor optimisation. You may have another approach in mind, but if not you may want to merge this; up to you, of course. :)

konsoletyper:
Why you are interested in contributing to TeaVM? Are you going to use it to translate some app? How did you find these issues?

davmac314:
I am one of the developers of Greenfoot (http://greenfoot.org), an IDE/animation framework for teaching and learning Java programming. Greenfoot allows uploading "scenarios" as Java applets, but the Java plugin is increasingly difficult to install/use as browsers do not want to support it and even Oracle appears to be deprecating it. Therefore, I am looking at possibilities for converting the Java scenarios to Javascript; right now, I am evaluating TeaVM with the idea of using it for this purpose. I found these problems while trying to re-implement the Greenfoot API using TeaVM's JSO wrappers.

konsoletyper:
Ok, you can email me (info@teavm.org), let's discuss it.

konsoletyper:
Ok, I think this optimization won't give much on the real code. Once I compared two approaches (boolean flag vs. hot patching) and it seems that first one is only slightly slower.

PR #246, opened 2017-02-21
Fix asynchronous class initialisation with asynchronous constructor

davmac314:
This fixes a bug which triggers when you have both asynchronous class initialisation (clinit) and asynchronous constructor. A test case is included.

### DIFF
--- a/core/src/main/java/org/teavm/backend/javascript/rendering/Renderer.java
+++ b/core/src/main/java/org/teavm/backend/javascript/rendering/Renderer.java
@@ -25,6 +25,7 @@ import org.teavm.ast.AsyncMethodNode;
 import org.teavm.ast.AsyncMethodPart;
 import org.teavm.ast.ClassNode;
 import org.teavm.ast.FieldNode;
+import org.teavm.ast.GotoPartStatement;
 import org.teavm.ast.MethodNode;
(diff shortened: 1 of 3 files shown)


davmac314:
Hmm. Something might be wrong if the test is failing. I'm investigating now.

davmac314:
Test definitely passes when I run them locally. The travis build check seems to have taken the second commit (the test itself) but not the fix? In any case the test is correct (and does fail previously), I think this is ok to merge.

core/src/main/java/org/teavm/backend/javascript/rendering/Renderer.java
```diff
@@ -658,6 +662,9 @@ public void visit(NativeMethodNode methodNode) {
         @Override
         public void visit(RegularMethodNode method) {
             try {
+                if (doClinit) {
+                    writer.appendClass(method.getReference().getClassName()).append("_$callClinit();").softNewLine();
```
konsoletyper: Can't this line be shared with `visit(NativeMethodNode methodNode)`?

davmac314:
Ok, that would probably be better. Do you want me to issue another pull request just with the test?

konsoletyper:
I can take single commit from this PR

PR #247, opened 2017-02-23
Fix asynchronous method detection

davmac314:
This fixes what looks like an obvious error in code, which makes a tool I have written (using the TeaVMTool API) produce correct code in certain cases. However it does also seem to prevent one of my Maven-based builds from working correctly.

I tried to create a test case for the problem but cannot seem to actually produce a test failure. The test case I've added at least shows exception output when run, but for some reason doesn't cause the test to fail. I think this might be an issue with the test runner. The fix I itself is I think "obviously correct" but perhaps I'm misunderstanding something in the code. Hope you can review. Thanks!


konsoletyper:
I pushed these two commits from CLI. Sorry, I had to modify your commits a little.

Issue #249, opened 2017-03-23
Throwable.getCause() causes Stack Overflow


shannah:
Calling `getCause()` on any `Throwable` object will cause a stack overflow.   The reason is that `getCause()` is rendered in javascript as:

~~~~
function jl_Throwable_getCause($this) {
    return jl_Throwable_getCause($this);
}
~~~~

(I.e. it just calls itself)

The Java version of this method is:

~~~~
@Override
    public TThrowable getCause() {
        return cause != this ? cause : null;
    }
~~~~

Which looks fine.

Test case at 
https://github.com/shannah/teavm-issue-248

(Same test case as for issue 248).  To cause the stack overflow, you can press the "Test Exception" button.

If I manually change the getCause() method in Javascript to

~~~~
function jl_Throwable_getCause($this) {
    return $this.$cause;
}
~~~~

It works fine.

PR #251, opened 2017-03-24
Fixes issue with stack overflow upon calling Throwable.getCause().  T…

shannah:
…here was likely another issue in the TeaVM compiler... this doesn't fix that.  It just changes the structure of the method, and TeaVM seems to be able to compile it properly now.

Ref.  https://github.com/konsoletyper/teavm/issues/249

This also changes printStackTrace to show the cause exceptions also. 
Also Throwable(Throwable cause) now sets the message to be the cause's message to bring it inline with OpenJDK's behaviour.


shannah:
Actually... in looking again, this doesn't fix the issue with Throwable.getCause() at all.  I had changed my test to just use printStackTrace() and forgot about it.  This change does show the cause in the stack trace though, which reduces the "need" for getCause().

Issue #258, opened 2017-04-05
WASM: Unreferenced but exported methods are removed by dependency checker


mirkosertic:
Consider the following code:

```
package de.mirkosertic;

import org.teavm.interop.Export;

public class Main {

    @Export(name = "passThru")
    public static int passThru(int aValue) {
       return aValue + 1;
    }

    public static void main(String[] args) {
    }
}
```

The passThru Method is removed by dependency checker, probably as it seems to be unused from the Java bytecode point of view. But the methods is intended to be called from JavaScript. 

Exported methods should always be part of the dependency graph.

konsoletyper:
It's an issue by design. TeaVM does not scan full classpath, since it can be large and imply big compilation time. It rather starts from known *main* class and then parses additonal classes when they are reached by dependencies. There must be a way to tell TeaVM that a certain method should be scanned. `@Export` annotation can't serve for this purpose, it's only purpose is to tell TeaVM to generate certain WebAssembly instructions.

mirkosertic:
I created pull request #259 as a proposal.

You are right, we cannot search the whole classpath for annotated methods. But we can search the already processed classes and for @Export annotated methods and mark them as used during dependency analysis. This mechanism would make the example code from above workable.

This is definitely a tradeoff between performance and usability, but it is better than nothing.

PR #259, opened 2017-04-05
#258 WASM: Unreferenced but exported methods are removed by dep checker

mirkosertic:
This is a proposal for using the @Export annotation to mark a method as an entry point,

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/generate/WasmDependencyListener.java
+++ b/core/src/main/java/org/teavm/backend/wasm/generate/WasmDependencyListener.java
@@ -15,9 +15,8 @@
  */
 package org.teavm.backend.wasm.generate;
 
-import java.util.ArrayList;
-import java.util.List;
-
+import java.util.HashSet;
+import java.util.Set;
 import org.teavm.dependency.AbstractDependencyListener;
 import org.teavm.dependency.DependencyAgent;
 import org.teavm.dependency.MethodDependency;
@@ -36,12 +35,10 @@ import org.teavm.model.MethodReader;
 
 public class WasmDependencyListener extends AbstractDependencyListener implements ClassHolderTransformer {
 
-    private List<MethodHolder> exportedMethods;
+    private final Set<String> visitedClassesCheckedForExport;
 
-    @Override
-    public void started(DependencyAgent agent) {
-        exportedMethods = new ArrayList<>();
-        super.started(agent);
+    public WasmDependencyListener() {
+        visitedClassesCheckedForExport = new HashSet<>();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


konsoletyper:
Does this PR really help?

core/src/main/java/org/teavm/backend/wasm/generate/WasmDependencyListener.java
```diff
@@ -55,6 +68,20 @@ public void transformClass(ClassHolder cls, ClassReaderSource innerSource, Diagn
...
+            AnnotationReader exportAnnot = method.getAnnotations().get(Export.class.getName());
+            if (exportAnnot != null) {
+                exportedMethods.add(method);
+
+            }
+        }
+    }
+
+    @Override
+    public void beforeCompleting(DependencyAgent agent) {
```
konsoletyper: You should be **very** careful when overriding `beforeCompleting`. This looks suspicious

konsoletyper: Here, you just link methods which were already reached (since `methodReached` was called for all of these methods).

mirkosertic:
It helps me, yes. See #258, it fixes the mentioned problem

core/src/main/java/org/teavm/backend/wasm/generate/WasmDependencyListener.java
mirkosertic: Well, methodReached is not called for @Export annotated methods. transformClass checks for @Export methods and adds them to dependency analysis by caching them. beforeCompleting is a new lifecycle method i introduced to allow additional stuff like this *after* dependency analysis was completed and *before* completion stage

konsoletyper:
Why did you add a new listener method? I suppose, it's enough to get `ClassReader` instance from withing `methodReached`, scan its methods and `link` exported ones? Did you try this approach?

mirkosertic:
I thought about that, but scanning the whole class for every reached method might be a little redundant, so I implemented this logic in transformClass, as this is only called one regardless how many reached methods a class has.

konsoletyper:
Isn't scanning a class is what you are *already* doing? Hint: in your dependency listener you may memoize classes you already scanned and don't scan them repeatedly.

mirkosertic:
so we would use agent.getClassSource().get(method.getReference().getClassName()).getMethods() within methodReached, crawl for annoted methods and mark them as used?

konsoletyper:
Exactly! More specifically:
```java
private Set<String> visitedClasses = new HashSet<>();

String className = method.getReference().getClassName();
if (visitedClasses.add(className)) {
    agent.getClassSource().get(className).getMethods();
    // etc
}
```

mirkosertic:
Wouldn't it be even better to use classReached instead?

mirkosertic:
Ok, so we'll use methodReached then?

konsoletyper:
I'm not sure. Let me investigate and think a bit. I just thought that `classReached` might be better from user's point of view.

mirkosertic:
In my example classReached works perfectly, but haven't tested it further

PR #272, opened 2017-04-16
Running Knockout for Java tests

jtulach:
I am converting the ko4j tests to JUnit's `@RunWith`. Right now I am getting an error, but the generated `test-js` files seem to be running OK. What shall I do to avoid the error?

```
  KnockoutTCKTest.loadAndParseJSONArray Class java.util.WeakHashMap was not found
    at org.teavm.junit.TeaVMTestRunner.<clinit>(TeaVMTestRunner.java:81)
    at java.lang.Class.forName(TClass.java:149)
    at java.lang.Class.forName(TClass.java:159)
    at org.netbeans.html.json.impl.JSON.initClass(JSON.java:453)
    at org.netbeans.html.json.impl.JSON.findType(JSON.java:387)
    at org.netbeans.html.json.impl.JSON.bindTo(JSON.java:396)
    at net.java.html.json.Models.bind(Models.java:83)
    at net.java.html.json.tests.JSONTest.loadAndParseJSONArray(JSONTest.java:549)
    at org.teavm.html4j.test.KnockoutTCKTest.loadAndParseJSONArray(KnockoutTCKTest.java:305)
    at org.teavm.junit.TestEntryPoint.launchTest
    at org.teavm.junit.TestEntryPoint.lambda$run$0(TestEntryPoint.java:27)
    at $$LAMBDA0$$.launch
    at org.teavm.testing.SimpleTestRunner.run(SimpleTestRunner.java:25)
    at org.teavm.junit.TestEntryPoint.run(TestEntryPoint.java:27)
```

It looks to me that the compilation believes that the `Class.forName` in `JSON.initClass` can load `TeaVMTestRunner`, but that is certainly not something that happens in the actual test...

### DIFF
--- a/html4j/src/test/java/org/teavm/html4j/test/KnockoutFXTest.java
+++ b/html4j/src/test/java/org/teavm/html4j/test/KnockoutFXTest.java
@@ -47,11 +47,16 @@ import org.testng.Assert;
 public final class KnockoutFXTest extends KnockoutTCK implements Transfer, WSTransfer<WSImpl> {
     private static Class<?> browserClass;
     private static Fn.Presenter browserContext;
+
     private KO4J ko4j = new KO4J();
     private final Map<String, Request> urlMap = new HashMap<>();
     private final Map<String, WSImpl> wsUrlMap = new HashMap<>();
 
     public KnockoutFXTest() {
+    }
+
+    static Class<?>[] allTestClasses() {
+        return testClasses();
     }
 
     static ClassLoader getClassLoader() throws InterruptedException {
--- a/html4j/src/test/java/org/teavm/html4j/test/KnockoutFXTest.java
+++ b/html4j/src/test/java/org/teavm/html4j/test/KnockoutFXTest.java
@@ -24,7 +24,6 @@ import java.util.HashMap;
 import java.util.Map;
 import net.java.html.BrwsrCtx;
 import net.java.html.js.JavaScriptBody;
-import org.netbeans.html.boot.spi.Fn;
 import org.netbeans.html.context.spi.Contexts;
 import org.netbeans.html.json.spi.JSONCall;
 import org.netbeans.html.json.spi.Technology;
@@ -38,16 +37,12 @@ import org.teavm.jso.JSObject;
 import org.teavm.jso.browser.Window;
 import org.teavm.jso.dom.html.HTMLDocument;
 import org.teavm.jso.dom.html.HTMLElement;
-import org.testng.Assert;
 
 /**
  *
  * @author Jaroslav Tulach <jtulach@netbeans.org>
  */
 public final class KnockoutFXTest extends KnockoutTCK implements Transfer, WSTransfer<WSImpl> {
-    private static Class<?> browserClass;
-    private static Fn.Presenter browserContext;
-
     private KO4J ko4j = new KO4J();
     private final Map<String, Request> urlMap = new HashMap<>();
     private final Map<String, WSImpl> wsUrlMap = new HashMap<>();
@@ -57,29 +52,6 @@ public final class KnockoutFXTest extends KnockoutTCK implements Transfer, WSTra
 
     static Class<?>[] allTestClasses() {
         return testClasses();
-    }
-
-    static ClassLoader getClassLoader() throws InterruptedException {
-        while (browserClass == null) {
-            KnockoutFXTest.class.wait();
-        }
-        return browserClass.getClassLoader();
-    }
-
-    public static void initialized(Class<?> browserCls) throws Exception {
-        browserClass = browserCls;
-        browserContext = Fn.activePresenter();
-        KnockoutFXTest.class.notifyAll();
-    }
-
-    public static void initialized() throws Exception {
-        Assert.assertSame(
-            KnockoutFXTest.class.getClassLoader(),
-            ClassLoader.getSystemClassLoader(),
-            "No special classloaders"
-        );
-        KnockoutFXTest.initialized(KnockoutFXTest.class);
-        browserContext = Fn.activePresenter();
     }
 
     @Override
--- a/html4j/src/test/java/org/teavm/html4j/test/KnockoutTCKCheckTest.java
+++ b/html4j/src/test/java/org/teavm/html4j/test/KnockoutTCKCheckTest.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2017 Jaroslav Tulach.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.teavm.html4j.test;
+
+import java.lang.reflect.Method;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+import org.junit.Test;
+import org.netbeans.html.json.tck.KOTest;
+
+/**
+ *
+ * @author Jaroslav Tulach
+ */
+public class KnockoutTCKCheckTest {
+    @Test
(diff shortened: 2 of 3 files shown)


konsoletyper:
You can eliminate call to `Class.forName`.

jtulach:
Can I eliminate call to `Class.forName`? No, I probably cannot - it is needed on some platforms and it was there for ages: https://github.com/jtulach/html-java-api/commit/febfe2f1f2f47ec653438e5dfa3af117149c2a4a - it's purpose is to make sure static initializer of the class has been invoked.

But maybe, on TeaVM, the whole method `JSON.initClass` could be intrinsified to no-op. How can the HTML4J support provide an intrinsic for the method? I see `kotlin.jvm.internal.Intrinsics`, I am not sure that is the right path to follow...

konsoletyper:
Ok, I'll try to intrisify this method

konsoletyper:
I could fix the problem, but the tests do not pass correctly in Rhino (though they do pass in browser). I think additional travis setup needed to run, say, in PhantomJS.

konsoletyper:
I just pushed my changes to master. You can rebase your branch and check. My suggestion is following: there should be an option to skip running JS tests (just remove corresponding system properties from surefire), then it's necessary to write a little crawler in JS that finds all JS files in given folder and runs them in PhantomJS. Can you do it yourself?

konsoletyper:
Ok, I created a new test runner that runs tests in Firefox. I'll provide explanation of how to run this runner manually a bit later. For now, I set this runner as default in travis configuration. Some of ko4j tests still fail. I'll investigate this a little later. As for my experience with former versions of ko4j there weren't any errors in html4j support, these were bugs in some mocks. I hope most of recent errors are because of bugs in test environment as well. So I suppose latest build of TeaVM (which will be hopefully available in few hours) is suitable for compiling DukeScript apps.

jtulach:
Sorry for being so slow. I've just merged the master changes in, let see what they will do.

jtulach:
Right, I also assume the tests to be failures in the testing environment (firefore, travis or mocks). Being able to reproduce the problems locally would help me to investigate more.

konsoletyper:
1. Run maven build with `-P \!htmlUnit` arguments, this should disable running tests in htmlUnit.
2. You should get all JS tests in `html4j/target/js-tests`. Each tests is represented as a folder which contains html file. You can simply open these html files and get test result/debug it in the devtools.

You can also rerun all tests automatically. You need node.js, not sure which version, try 6 or 7.
1. Go to `tests/src/test/js`
2. Run `npm install && npm run build`
3. Open `index.html` in the browser
4. Run `node start.js <path-to-test-folder>`

BTW, I noticed that many tests fail in `optimized` variant. This may be due to bugs in TeaVM optimizer, I'll investigate these tests myself.

konsoletyper:
It's gonna be challenging. I don't know codebase of ko4j. All these bindings, context, techs, etc, it seems overcomplicated to me. On the other side, you may find it hard to debug code with most of the functions inlined in one huge single function. I don't know that to do.

jtulach:
It surprises me that when I execute
```
$ mvn -f html4j clean install -P \!htmlUnit
...
[INFO] BUILD SUCCESS
```
it prints success. Shouldn't the build fail? But yeah, I managed to get over it to running the tests via node per your instructions. At the end I just went into `html4j/target/js-tests/org/teavm/html4j/test/KnockoutTCKTest/rawObject`, opened the `run-test.html` where I replaced reference from `test.js` to `test-optimized.js` and the failure became visible.

Very suspicious thing in the optimized run is that I see usage of `onhji_JSON$EmptyTech` - that would signal some `ServiceLoader` lookup failed and the system isn't properly initialized as the proper technology should be `onhk_KOTech` as in the non-optimized case. When running the non-optimized case the `KOTech` constructor is called via:
```
onhk_KOTech__init_ (test.js:8683)
onhk_KO4J_knockout (test.js:18246)
otht_KnockoutFXTest_createContext (test.js:18551)
njhjt_Utils_newContext (test.js:18812)
njhjt_KnockoutTest_newContext (test.js:7684)
njhjt_KnockoutTest_rawObject (test.js:7637)
otht_KnockoutTCKTest_rawObject (test.js:20163)
otj_TestEntryPoint_launchTest (test.js:14031)
otj_TestEntryPoint_lambda$run$0 (test.js:14034)
$$LAMBDA0$$_launch (test.js:20891)
ott_SimpleTestRunner_run (test.js:15761)
otj_TestEntryPoint_run (test.js:14022)
(anonymous) (run-test.html:24)
TeaVMThread.run (runtime.js:545)
TeaVMThread.start (runtime.js:532)
$rt_startThread (runtime.js:574)
(anonymous) (run-test.html:11)
```
The constructor is not called in the optimized case at all. I'd say there is some bug in the optimizer.

konsoletyper:
> Shouldn't the build fail?

It should not. This profile disables running tests in htmlunit, therefore only JS files built (which are incorrect). Later, you can run tests via node.js and see what really fails.

> onhji_JSON$EmptyTech - that would signal some ServiceLoader lookup failed and the system isn't properly initialized as the proper technology should be onhk_KOTech as in the non-optimized case.

Ok, this should be a useful hint. I'll check whether devirtualization is performed correctly. I suspect it devirtualized call site that should be virtual, therefore you get (wrong) statically resolved class.

> I'd say there is some bug in the optimizer.

Sure, I noticed that most of the tests fail only in optimized version, the regular versions run without errors. This is possible only due to some bug in optimizer.

konsoletyper:
I think I found at least one issue. Here's the code somewhere in KO4J TCK:

```java
    @Override
    public BrwsrCtx createContext() {
        KO4J ko4j = new KO4J();
        return Contexts.newBuilder()
                .register(Transfer.class, this, 1)
                .register(Technology.class, ko4j.knockout(), 1) // <-- !! Here
                .register(WSTransfer.class, this, 1)
                .build();
    }
```

The line I marked gets translated to following code:

```js
$ko4j$ko4j = null;
// some code here
var$3 = $rt_cls(onhjs_Technology);
    var$4 = onhcs_Contexts$Builder_register(onhcs_Contexts$Builder_register(var$4, var$3, $ko4j$ko4j, 1), $rt_cls(onhjs_WSTransfer), $this, 1).$impl;
```

What happens here is

1. `KO4J.<init>` method is inlined
2. `KO4J.knockout` method is inlined
3. Since `ko4j` instance does not escape, TeaVM applies scalar replacement; `ko4j` is replaced with several local variables representing its fields, including `ko4j$ko4j`.
4. Then, constant condition elimination sees that `ko4j$ko4j` is always `null`, thus it's possible to eliminate dead branch.

Looks like the last step has implementation bug which leads to removal of *wrong* branch. That's why `ko4j$ko4j` is not fed with actual implementation of knockout.

konsoletyper:
Further investigation shown that optimization removed correct branch, but failed to maintain phi functions in corner case when one branch directly jumps to another branch (i.e `if` without `else`). I patched optimizer and now more tests pass. I'll push my fix soon.

konsoletyper:
Ok, I pushed the fix, and now only 17 tests fail, including 5 tests in optimized versions, and remaining 4x3 in all versions.

konsoletyper:
Ok, I fixed bug that caused remaining 5 tests fail with optimizations. Now only 4 tests fail, I expect due to bug in html4j implementation. Can you help me to debug them? Here is the list: deserializeWrongEnum, displayContentOfAsyncArray, modifyValueAssertAsyncChangeInModel, paintTheGridOnClick

jtulach:
Hello Alexey, the `deserializeWrongEnum` failure is caused by TeaVM allowing me to call `System.setErr`, but then ignoring my `OutputStream` when I do `ex.printStrackTrace()`. If you throw `SecurityException` in `System.setErr`, the test shall pass.

I am looking at the other failures.

konsoletyper:
Ok, I fixed the issue with System.setErr/setOut, and now all tests pass, except one that does not even compile. It's `triggerEvent`. Did you add it by a mistake? Tests with parameters are not supported by TeaVM test runner.

jtulach:
Thanks. When do you plan to release next version of TeaVM? I am eager to
announce it is HTML/Java API 1.4 compliant...

2017-05-08 13:40 GMT+02:00 Alexey Andreev <notifications@github.com>:

konsoletyper:
I don't know, I don't have any schedule. I can release development build at the end of this week.

konsoletyper:
I published development build of TeaVM on bintray here: https://bintray.com/konsoletyper/teavm/teavm-dev/0.5.0-dev-328

I don't call it "release", but it's stable and reliable enough and can be used to write application with html4j/DukeScript.

PR #273, opened 2017-04-19
Full HTML5 api support for TeaVM JSO

liraz:


liraz:
Hey @konsoletyper, i did full HTML5 JSO api for teavm, could you check it please?
Seems like the rebase made a lot of other commits come in.

konsoletyper:
Sorry, I can't accept this review. First, you haven't even discussed your changes with me before creating PR. Second, it's hard to review 106 commits with completely unrelated changes

liraz:
@konsoletyper 

> First, you haven't even discussed your changes with me before creating PR.

I added JSO wrappers for all HTML5 api - if this is not needed please tell me now - i will just close this PR.
If you do think it is needed, i will recreate it with only relevant commits - thanks.

konsoletyper:
HTML5 APIs indeed are needed. But I don't like how you wrote these APIs.

liraz:
Have you ever even looked at my code? it is very similar to the wrappers you did - and to tell you the truth i spent quite a while on this code, and this is the thanks i get.. then i don't wanna help this project at all.

konsoletyper:
Yes, I did. What I really didn't like is code like following:

```java
    default EventPhase getEventPhase() {
        return UseHTMLValue.toEnumValue(EventPhase.class, innerGetEventPhase());
    }

    @JSProperty("eventPhase")
    int innerGetEventPhase();
```

Here, you `innerGetEventPhase`, which is definitely not intended to be public API; however it's public. Instead of mapping enums in the wrappers, it's better to add support of mapping enums to JSO. 

Another example is:

```
<E extends HTMLElement> HTMLCollection<E> getElements()
```

For which purpose `E` parameter is here?

Did not you receive messages from me before? It's strange, my be it was something wrong with GitHub. Anyway, your commit it quite large, I have to look though it carefully again and leave my concerns there. You better discuss such big changes with me before you are going to make them.

liraz:
> Instead of mapping enums in the wrappers, it's better to add support of mapping enums to JSO.
There's no support for enum wrappers - actually saw the issue open on it, and i would be happy to give a shot for an implementation there (or it's already somewhat implemented)... first i need to calm down - kinda pissed at you.

> For which purpose E parameter is here?
It had a purpose before for using Element instead of HTMLElement, but kotlin example seem to fail because of it, so i just changed it back to HTMLElement for now.

Also there's no possible way of using these API in test cases - maybe now it's not an issue - i didn't check latest changes yet.

PR #274, opened 2017-04-27
Fixed hashCode() and compare Handling for TLong

mirkosertic:


konsoletyper:
Fixed myself, with better performance of the generated code

PR #281, opened 2017-05-28
java.util.Base64 support

lbrasseur:
java.util.Base64 downloaded and modified from here:

http://grepcode.com/file_/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/Base64.java/?v=source


konsoletyper:
I can't accept code from OpenJDK without legal issues.

lbrasseur:
GNI classpath would be a problem too? Looking into Apache Harmony would be better?

lbrasseur:
I meant, GNU Classpath

konsoletyper:
Apache Harmony is under Apache license, which is ok (TeaVM is also under Apache 2.0). I guess GNU classpath is under (L)GPL, which is incompatible with Apache 2.0.

lbrasseur:
Thanks! I will search there

PR #282, opened 2017-05-28
java.util.Base64 partial support

lbrasseur:
This is a partial implementation. Apache Harmony don't have a java.util.Base64 implementation (it is available since Java 8 and Apache Harmony implemented up to Java 6).

I built a partial solution using code from [Apache Harmony Base64](https://github.com/apache/harmony/blob/java6/drlvm/vm/vmcore/src/kernel_classes/javasrc/org/apache/harmony/misc/Base64.java)


lbrasseur:
I will remove the "throw new UnsupportedOperationException(":(");" methods because they will make the implemented methods report to fail.

Issue #279, opened 2017-05-25
JSFunctor defender methods


marciodel:
The JFunctor Wiki documentation states:

> Functor interfaces must contain exactly one method

It would be nice to support interfaces marked with JSFunctor with defender methods as overlays.

Thanks!

marciodel:
Sorry. It looks like it is supported, but it didn't work in a previous test. I'll close this issue and try to find the exactly condition that generates the bug.

Thanks!

PR #283, opened 2017-05-29
Support for default and static methods in Functors

marciodel:
This pull request implements the enhancement #279 .


marciodel:
@konsoletyper, would you mind to have a look? It seems that the build problems are not related with the Pull request.

konsoletyper:
Don't create pull request for 0.4.x. Create for `master` instead.

marciodel:
Ok. I'll try. I have an issue where I not managed to use the master version with the gradle plugin. I've changed the gradle plugin to use master version, but it complains about some missing libraries. I'll try to fix that.
Thanks!

PR #286, opened 2017-06-17
java.util.ResourceBundle partial implementation

lbrasseur:
This is a partial implementation for java.util.ResourceBundle. The base code was taken from Apache Harmony.


konsoletyper:
Can you provide tests?

lbrasseur:
I tried to port the tests from Apache Harmony but I had problems mocking TClassLoader and with TLocale  (it is also initialized in static code). I will do some research to see if there is some solution to this.

ScraM-Team:
@lbrasseur , does this allow properties files to be used for localization?  Do you have an example or sample project you could share?

(@vic-cw is interested too)

lbrasseur:
Oh, this is just a POC I did some time ago. I think it is not working. Sorry for not being helpful :(

vic-cw:
Thanks for including me @ScraM-Team

PR #300, opened 2017-08-28
JSO Websocket implementation.

adamjryan:


konsoletyper:
Please, reformat your code to conform TeaVM code style. You can learn the code style by examining other sources. Note that code style is enforced by checkstyle and project won't even compile if the source code does not conform code style (as shown by travis build).

jso/apis/src/main/java/org/teavm/jso/websocket/MessageEvent.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.teavm.jso.JSBody;
+import org.teavm.jso.JSObject;
+import org.teavm.jso.dom.events.Event;
+import org.teavm.jso.typedarrays.ArrayBuffer;
+
+public abstract class MessageEvent implements Event
+{
+    /** @return the data sent by the message emitter. */
+    @JSBody( script = "return this.data;" )
```
konsoletyper: Another possible way to express this is:

```java
@JsProperty
public abstract JsObject getData();
```

jso/apis/src/main/java/org/teavm/jso/websocket/WebSocket.java
```diff
@@ -0,0 +1,163 @@
...
+    private static final Logger LOGGER = Logger.getLogger( WebSocket.class.getName() );
+
+    public static void newInstance( String url, NewWebSocketCallback callback )
+    {
+        newInstance( url, null, callback );
+    }
+
+    public static void newInstance( String url, String[] protocols, NewWebSocketCallback callback )
+    {
+        LOGGER.config("Creating new instance.");
```
konsoletyper: I'm not sure logging is really necessary here

adamjryan:
Thanks for the feedback. It was from a pet project, although I'm working on a different bit now so will look to push up some updates when I get back to playing with it again.

jso/apis/src/main/java/org/teavm/jso/websocket/MessageEvent.java
adamjryan: I think it better reflects in the api to just have the property name for the method name when it represents access to an exposed property on the javascript object. Are you happy with this?

adamjryan:
I shifted the extended Java methods of the Websocket overlay into a wrapping class (JWebSocket) which is currently in the same package as the overlay class. To keep the JSO package clean, where's the best place to put it?

konsoletyper:
> To keep the JSO package clean, where's the best place to put it?
I don't know. Currently, there are no convenient wrapper classes for JSO in TeaVM repository.

adamjryan:
This is ready to commit.

konsoletyper:
If you want this pull request to be accepted, you better fix things I pointed to.

adamjryan:
Some comments I have addressed, others I have left replies. What are you referring to?
There are two points that would be good to address and I'll reiterate here:
* The Async annotation changes the underlying program flow in a way that the user of the API would perhaps not expect. It's better to keep the library code simple and let the user control the use of the annotation in their own code, that way they are knowledgeable of it's implications if something goes awry in their application.
* The native property accessors present more nicely in my mind without the get/set prefix as they represent an access to a raw property on the JS object and not a method call mapping. That is why the @JsProperty annotation was not used as it needs the method name to be a get/set in order to generate the mapping. Agree/disagree?

konsoletyper:
I mean comments like this one

> Why not TITLECASE?

in `BinaryType.java`. Can you see it?

> The Async annotation changes the underlying program flow in a way that the user of the API would perhaps not expect. It's better to keep the library code simple and let the user control the use of the annotation in their own code, that way they are knowledgeable of it's implications if something goes awry in their application.

That's why you can provide *both* APIs. And I think there's misunderstanding about `Async` annotation. Compiler produces errors when Async API can't be called from a certain point of program. For example, from method annotated with `Sync` annotation, or from a JSO callback.

> The native property accessors present more nicely in my mind without the get/set prefix as they represent an access to a raw property on the JS object and not a method call mapping. That is why the @jsproperty annotation was used a field specifier as otherwise it needs the method name to be a get/set in order to generate the mapping. Agree/disagree?

I don't know. There's tons of opinions concerning this. Yes, methods without `get` prefix are 3 characters shorter. On the other hand, there are style guides that requre methods to be denoted by a verb, so method like `source` are improper, while proper ones are `getSource`. Another concern is: there is Java beans convention, that prepend name of getters with `get` prefix, this convention is clear to most of Java developers. Finally, there are existing APIs in JSO packages, whey follow some convention. Can you agree that APIs should be consistent?

If you still want to discuss, let me give you another point. I'm not sure JSO still forces you to call getters this way. I think I made a change that allows to call `JSProperty` methods as you wish, and the role of a particular method is determined by its signature. Please, try this

```java
@JSProperty
public abstract JSObject source();
```

and let me know if this does not work.

adamjryan:
Unfortunately I rebased my changes before and it meant the comments you added are against a commit that is no longer referenced. I didn't realise that would happened at the time, apologies.
I've made the changes you mentioned above, if there are any others please could you add the comments again.
It does make sense for consistency in the project to prefix the property methods get/set so they're updated. The Async annotation requires a adding a dependency on the interop module which doesn't seem like a good idea, perhaps there's a better place for native extension code.

automenta:
here is a simple 1-class TeaVM Client WebSocket:
https://gist.github.com/automenta/7038c43939adf75e604688d8aca832bd

it includes a small minified ReconnectingWebsocket wrapper as a String which is meant to be delivered to client by server via a separate http request. i just include it in the client code there for convenience.

automenta:
also my use of StringBuilder there was necessary for JDK11 because of it uses a String concatenation implementatoin that TeaVM doesnt recognize (yet).  but StringBuilder seems to works fine so far instead of +

PR #313, opened 2017-11-05
Added some missing dom methods and properties

niloc132:
These are broken into distinct commits, let me know if you want to get them merged one at a time.


jso/apis/src/main/java/org/teavm/jso/dom/xml/DOMTokenList.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.teavm.jso.dom.xml;
+
+import org.teavm.jso.core.JSArrayReader;
+import org.teavm.jso.core.JSString;
+
+public interface DOMTokenList extends JSArrayReader<JSString> {
+    boolean contains(String token);
+
+    void add(String... tokens);
```
konsoletyper: I'm not sure if varargs are supported in JSO. Please check. If they aren't, let's think how we can work-around this. And don't forget to create an issue to support vararg.

niloc132: A good question - I'll verify more thoroughly, but I can verify that calling `add(String...)` with a single argument does work as expected presently.

PR #315, opened 2017-12-01
Fixed NPE when padding null arrays to JSO methods that accept arrays.

shannah:
`JS.wrap((byte[])null)` and the like previously would throw NPE.  It makes sense that these should simply return `null` instead.  This pull request guards against null parameters in JS.wrap() and JS.unwrap, returning null in those cases.


konsoletyper:
Ok, but please, provide tests next time

PR #324, opened 2018-02-07
Added methods to expose the scrollY value and body offset

scxrx-xx:
As discussed by e-mail. I think this should be the appropriate place to put the methods. But I think I am missing something, since I cannot use them in my TeaVM client project. My IDE allows me to use these methods, but the maven build fails when I try to build my project.

This is the output from my build:

nl/scuro/fridge/client/collections/CollectionsView.java:[68,70] cannot find symbol
  symbol:   method scrollY()
  location: class org.teavm.jso.browser.Window
nl/scuro/fridge/client/collections/CollectionsView.java:[68,125] cannot find symbol
  symbol:   method getOffsetHeight()
  location: interface org.teavm.jso.dom.html.HTMLBodyElement

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/browser/Window.java
+++ b/jso/apis/src/main/java/org/teavm/jso/browser/Window.java
@@ -67,7 +67,7 @@ public abstract class Window implements JSObject, WindowEventTarget, StorageProv
     public abstract int getOuterHeight();
 
     @JSProperty
-    public abstract int scrollY();
+    public abstract int getScrollY();
 
     @JSProperty
     public abstract String getName();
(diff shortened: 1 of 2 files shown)


jso/apis/src/main/java/org/teavm/jso/browser/Window.java
```diff
@@ -66,6 +66,9 @@ private Window() {
     @JSProperty
     public abstract int getOuterHeight();
 
+    @JSProperty
+    public abstract int scrollY();
```
konsoletyper: should bet `getScrollY()`

scxrx-xx:
Yes, I've double checked everything. Should have checked it three times :) But it's working now. I've built the TeaVM parent project with the "with-cli" profile enabled, and the error disappeared.

PR #325, opened 2018-02-11
Minor fix for TeaVMRunner 'generate source maps' option

scxrx-xx:
The option to generate source maps was being configured with 'G', but checked for 'S' being passed as an option.


konsoletyper:
No, it's wrong change. `-G` it meant to be option to enable generation of source maps. Instead, you should fix code that checks `-S` option and change it to `-G`.

scxrx-xx:
Ok, I see. Changed it now.

PR #332, opened 2018-02-18
TPrintStream: add print(boolean) / println(boolean)

belowm:


classlib/src/main/java/org/teavm/classlib/java/util/concurrent/atomic/TAtomicReference.java
```diff
@@ -0,0 +1,56 @@
...
+import org.teavm.classlib.java.util.TObjects;
+
+/**
+ * @author martin below
+ */
+public class TAtomicReference<T> implements Serializable {
+    private T value;
+
+    public final boolean compareAndSet(T expect, final T update) {
+        if (!TObjects.equals(value, expect)) {
```
konsoletyper: According to [documentation](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicReference.html#compareAndSet(V,%20V)), referential comparison should be there.

classlib/src/main/java/org/teavm/classlib/java/util/concurrent/atomic/TAtomicLong.java
```diff
@@ -0,0 +1,135 @@
...
+    public final long decrementAndGet() {
+        return --value;
+    }
+
+    public final long addAndGet(final long delta) {
+        value += delta;
+        return value;
+    }
+
+    public final long getAndUpdate(final LongUnaryOperator updateFunction) {
```
konsoletyper: FYI. You must have copy-pasted this code from my `TAtomicInteger` implementation. However, both this code and corresponding code in `TAtomicInteger` is wrong. Actually, TeaVM is not single-threaded, context switch can occur as you call any method, and thus this code should be synchronized. I'll accept this code anyway and try to fix it later myself.

classlib/src/main/java/org/teavm/classlib/java/util/concurrent/atomic/TAtomicReference.java
konsoletyper: Also, this code won't work property in multi-threaded environment, see my notes to `TAtomicLong`.

konsoletyper:
@belowm did you receive my review?

Issue #334, opened 2018-04-08
NPE in ReflectClassImpl.getFields()


sheepdreamofandroids:
Wow! I'm really impressed by this project! I love the simplicity of the programming model and the speed and efficiency of the compiler.
While experimenting with the metaprogramming api (as a GWT refugee I'm trying to recreate serialization...) I ran into the aforementioned problem.
It happens on line 340 ( the middle one in below fragment) where classReader is null. I see you take great care of that in other methods but not here:

```
            fieldsCache = context.getClassSource()
                    .getAncestors(classReader.getName())
                    .flatMap(cls -> cls.getFields().stream().filter(fld -> fld.getLevel() == AccessLevel.PUBLIC))

```
I found that it is easily possible to debug the compiler by debugging maven. Would debugging be possible through the plugin?

Cheers, Guus

konsoletyper:
> Would debugging be possible through the plugin?

You mean debugging IDEA plugin? Ctrl+Shift+A -> Debug build process. Then, when you press 'build' button, IDEA starts build process in debug mode and shows remote port. You can create remote debug launch configuration with corresponding port to start debugging. Additionally, you can press Ctrl+Shift+A -> Registry... and set `compiler.process.debug.port` value to, say, 5005.

Personally, I debug CLI process. See `teavm-cli` module, `TeaVMRunner` class.

konsoletyper:
You don't need to recreate serialization and RPC. There are [JSON](http://teavm.org/docs/flavour/json.html) and [REST](http://teavm.org/docs/flavour/rest-client.html) support in TeaVM.

sheepdreamofandroids:
I didn't know about 'Debug build process'. Nifty! But I can see how debugging teavm-cli is even handier. I'll try that.

I know about JSON but I'm not a fan. It's too limited and undefined to my taste (no shared structure, no difference between lists/sets, no polymorphism). On top of that it is just inefficient. Most of that can be worked around but why bother?

Anyway, what I really want is to synchronize structures between client and server, preferably using websockets so REST is really not a concern.

konsoletyper:
> no shared structure

What do you mean?

> no difference between lists/sets

There's no need in it. To *transfer* data, you need only transfer items. If these items are stores in either `Set<T>` or `List<T>` in Java, they'll be just put there.

> no polymorphism

[...]

PR #335, opened 2018-04-08
JIRA #334 - check for null classreader

sheepdreamofandroids:
Just a small fix for #334


konsoletyper:
Instead of checking for `null`, I'd investigated *why* `classReader` is `null`.

sheepdreamofandroids:
classReader was null because resolve() wasn't called first. I guess my commit comment was misleading.

PR #338, opened 2018-05-04
Added TIdentityHashMap

comick:
It is a copy of `THashMap` using `identityHashCode` and reference equality.
Runtime has been changed to have a better distribution for objects ids returned by `identityHashCode`.

### DIFF
--- a/core/src/main/resources/org/teavm/backend/javascript/runtime.js
+++ b/core/src/main/resources/org/teavm/backend/javascript/runtime.js
@@ -15,15 +15,14 @@
  */
 "use strict";
 var $rt_global = this;
-var $rt_lastObjectId = 1;
+var $rt_seed = Math.floor(Math.random() * (0x7fffffff + 0x80000000 + 1)) - 0x80000000;
 function $rt_nextId() {
-    var current = $rt_lastObjectId;
-    var next = (current + 1) | 0;
-    if (next === 0) {
-        next = (next + 1) | 0;
-    }
-    $rt_lastObjectId = next;
-    return current;
+    var x = $rt_seed;
+    x ^= x << 13;
+    x ^= x >> 17;
+    x ^= x << 5;
+    $rt_seed = x;
+    return x;
 }
 function $rt_compare(a, b) {
     return a > b ? 1 : a < b ? -1 : 0;
--- a/core/src/main/resources/org/teavm/backend/javascript/runtime.js
+++ b/core/src/main/resources/org/teavm/backend/javascript/runtime.js
@@ -15,7 +15,7 @@
  */
 "use strict";
 var $rt_global = this;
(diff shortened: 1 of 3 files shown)


core/src/main/resources/org/teavm/backend/javascript/runtime.js
```diff
@@ -15,15 +15,10 @@
...
-    var current = $rt_lastObjectId;
-    var next = (current + 1) | 0;
-    if (next === 0) {
-        next = (next + 1) | 0;
-    }
-    $rt_lastObjectId = next;
-    return current;
+    var min = -0x80000000;
+    var max = 0x7fffffff;
+    return Math.floor(Math.random() * (max - min + 1)) + min;
```
konsoletyper: I don't like this. It should affect performance, right? Perhaps there should be better approach (i.e. rotate by fixed ratio and xor by fixed value).

comick:
Using original xorshift approach from Marsaglia paper "Xorshift RNGs" [¹] with seed generated using `Math.random`. There are more sophisticated ways, don't think they are needed in this case.
Roughly 3x faster than previous approach. Any feedback on that? Tests will come.

[1] http://www.jstatsoft.org/v08/i14/paper

konsoletyper:
As for performance. Do you have any benchmark that shows performance degradation with incrementing id generator? Does this benchmark show any boost after changing id generation algorithm? What is the performance of xorshift compared to the simple increment?

comick:
Added few very specific tests. Regarding _raw_ id generation performance I ran this snippet in Firefox 59 and Chromium 65.
```javascript

var $rt_lastObjectId = 1;
function $rt_nextIdIncr() {
    var current = $rt_lastObjectId;
    var next = (current + 1) | 0;
    if (next === 0) {
        next = (next + 1) | 0;
    }
    $rt_lastObjectId = next;
    return current;
}

var $rt_seed = 2463534242;
function $rt_nextId() {
    var x = $rt_seed;
    x ^= x << 13;
    x ^= x >> 17;
    x ^= x << 5;
    $rt_seed = x;
    return x;
}

for (var i = 0; i < 20; i++) {
    var times = 100000000;
    console.time('xorshift')
    for (var j = 0; j < times; j++) {
        $rt_nextId();
    }
    console.timeEnd('xorshift')

    console.time('incr')
    for (var j = 0; j < times; j++) {
        $rt_nextIdIncr();
    }
    console.timeEnd('incr')
}
```

After warmup I got a median:
firefox (xorshift ~1.6x slower than increment):
```
xorshift: 444ms
incr: 268ms
```
chromium (xorshift ~1.8x slower than increment):
```
xorshift: 573.159912109375ms
incr: 321.7890625ms
```

In this content performance is about a balance between raw unique id generation and performance improvement of `TIdentityHashMap` due to improved distribution of ids.

What do you think?

PR #341, opened 2018-05-17
Fix TSetFromMap#add return value

comick:
Noticed while trying to use `TCollections.newSetFromMap`. The fix makes it fit the `Collections#add` documentation stating: `@return <tt>true</tt> if this collection changed as a result of the call`


comick:
I build is green and the change is accepted I will add some testing as well.

PR #343, opened 2018-06-03
Added offsetTop and offsetLeft DOM methods

jupake:


jupake:
All tests pass when I run the test suite manually. Not sure why this failed.

konsoletyper:
Tests may fail due to different reasons on the build server. It may be build server failure. Or, as far as I can see, IntelliJ distribution of a certain version (used as a dependency) was removed. You should examine Travis logs to find the problem. Anyway, you changes are not involved.

PR #358, opened 2018-09-14
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/konsoletyper/teavm/ci/#badges) to your project. They will indicate a good code quality to potential users and contributors.
To get an idea of the analyses reflected by these grades, check the [alerts discovered by LGTM](https://lgtm.com/projects/g/konsoletyper/teavm).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ TeaVM
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.teavm/teavm-maven-plugin/badge.svg)](https://maven-badges.herokuapp.com/maven-central/org.teavm/teavm-maven-plugin) 
 [![Bintray](https://api.bintray.com/packages/konsoletyper/teavm/teavm-dev/images/download.svg)](https://bintray.com/konsoletyper/teavm/teavm-dev/_latestVersion)
 [![Gitter chat](https://img.shields.io/badge/gitter-join%20chat-green.svg)](https://gitter.im/teavm/Lobby)
+[![Code Quality: Java](https://img.shields.io/lgtm/grade/java/g/konsoletyper/teavm.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/konsoletyper/teavm/context:java)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/konsoletyper/teavm.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/konsoletyper/teavm/alerts)
 
 See documentation at the [project web site](http://teavm.org/).
 


konsoletyper:
Here's my feedback:

1. What is the purpose on integrating such service in GitHub? I don't use GitHub to write code, I use IntelliJ IDEA instead and want code quality inspections there.
2. Is there any way to suppress inspections in places where I'm sure on what I'm doing?
3. 'Implicit cast of source type long to narrower destination type int.' reported. But this was my intention to get 'int' and widen it to 'long' type, according to Java spec.
4. `MetaprogrammingClassLoader.java:46` is wrongly reported. `BufferedInputStream` wraps actual native stream, which may hold native resources and which is correctly closed by try-with-resources block.

xcorail:
Hi @konsoletyper 

Thank a lot for the feedback!

> What is the purpose on integrating such service in GitHub? I don't use GitHub to write code, I use IntelliJ IDEA instead and want code quality inspections there

Most code inspections are pretty fast and should then happen at the earliest time possible, when writing code, in your IDE, I agree. But most analysis done by LGTM are quite deeper than these simple code linters. They use control flow and tainting mechanisms that require time and memory, and these analysis cannot happen in your IDE, it would slow you down instead of helping you. We think that given the time taken by these analysis, the pull request / code review is the most appropriate place to run them. Look at examples of such queries on the [blog](https://lgtm.com/blog), for example [this one](https://lgtm.com/blog/apache_struts_CVE-2018-11776). Our queries are open sourced at https://github.com/Semmle/ql 

> Is there any way to suppress inspections in places where I'm sure on what I'm doing?

Sure. By putting a `//lgtm` in your code, as explained [here](https://lgtm.com/help/lgtm/alert-suppression). You will have also this information by clicking on the icon with the crossed eye at the right of the alert (red rectangle). 

>  'Implicit cast of source type long to narrower destination type int.' reported. But this was my intention to get 'int' and widen it to 'long' type, according to Java spec.

hmmm ... the alert here just says that `pos += amount` will implicitly narrow long `amount` to int, possibly losing information. But in your case,  you know that `amount` will always be an int. So right, you might want to either suppress this alert (severity = warning, btw) because you know what you're doing here, or you can widen `pos` and `count` to long, to be consistent with what you did with `amount`.   

> MetaprogrammingClassLoader.java:46 is wrongly reported. BufferedInputStream wraps actual native stream, which may hold native resources and which is correctly closed by try-with-resources block

With a quick look, the `InputStream input` will be indeed closed in case of error thanks to the try-with-resources, but the `BufferedInputStream` will not, don't you agree? I will ask my Java experts to have a closer look.

Cheers, thanks again for the feedback.

ScraM-Team:
@xcorail , can your tool's analyses be performed offline, from the IDE or as part of a local build, or can they only be run via your service remotely against a public repo?  Tools that can't be run locally aren't very developer-friendly since the build can break unexpectedly after a commit.

xcorail:
Hi @ScraM-Team 

As I said in my previous answer, the analysis we perform are very expensive in terms of time and memory, and do not easily fit at commit time in a developer workflow. It would slow down instead of helping. You can make an analogy with unit tests, fast enough so that you can run them at build time, and more expensive integration or end to end tests, that you will launch only nightly or at pull request time: You are launching simple syntactic analysis in your IDE, simple linters at build or commit time, but you will launch the LGTM deep analysis at a later stage.

However, technically yes, the tool can be launched offline, with our [QL for eclipse plugin](https://lgtm.com/blog/ql_for_eclipse). But the intended usage is rather when you are working on **one** particular problem and you want to explore variants of this problem in your code, but not launching **all** the queries set, as this would consume all your resources and take too much time.
 
Hope this makes sense.

PR #382, opened 2019-03-18
Fix IllegalStateException in TDeflater

Quantum64:
TDeflater passes in strategy as the second parameter in jzlib's Deflater class.

https://github.com/ymnk/jzlib/blob/master/src/main/java/com/jcraft/jzlib/Deflater.java#L78

The second parameter is actually for window bits and a value of 0 causes the compressor initialization to fail. Simply removing this parameter fixes the issue.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/zip/TDeflater.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/zip/TDeflater.java
@@ -57,7 +57,7 @@ public class TDeflater {
         }
         compressLevel = level;
         try {
-            impl = new Deflater(compressLevel, strategy, noHeader);
+            impl = new Deflater(compressLevel, noHeader);
         } catch (GZIPException e) {
             // do nothing
         }


ScraM-Team:
What were you doing that caused this exception?  I haven't seen it yet.

Quantum64:
```
import java.util.zip.Deflater;

public class TestDeflate {
		Deflater def = new Deflater();
		def.setInput(new byte[] { 0 }, 0, 0);
	}
}
```
This should cause the issue. setInput throws the exception if Deflater fails to initialize, which it always does since currently the wrong parameters are passed to it.

The actual code I was compiling is
https://github.com/leonbloy/pngj
Where the issue becomes a problem here
https://github.com/leonbloy/pngj/blob/master/src/main/java/ar/com/hjg/pngj/pixels/CompressorStreamDeflater.java#L44

Quantum64:
Specifically, TeaVM passes in 0 for the second parameter of Deflater's constructor:
https://github.com/konsoletyper/teavm/blob/master/classlib/src/main/java/org/teavm/classlib/java/util/zip/TDeflater.java#L28
https://github.com/konsoletyper/teavm/blob/master/classlib/src/main/java/org/teavm/classlib/java/util/zip/TDeflater.java#L60

That gets passed to Deflater's init method which returns an error value since 0 isn't a valid option for this parameter:
https://github.com/ymnk/jzlib/blob/master/src/main/java/com/jcraft/jzlib/Deflater.java#L110

Then Deflater throws a GZIPException which is ignored by TeaVM:
https://github.com/ymnk/jzlib/blob/master/src/main/java/com/jcraft/jzlib/Deflater.java#L82
https://github.com/konsoletyper/teavm/blob/master/classlib/src/main/java/org/teavm/classlib/java/util/zip/TDeflater.java#L62

Then when setInput is called since the initialization failed that causes an IllegalStateException:
https://github.com/konsoletyper/teavm/blob/master/classlib/src/main/java/org/teavm/classlib/java/util/zip/TDeflater.java#L178

ScraM-Team:
Interesting.  The ScraM UI (based on Flavour and TeaVM, of course!) will eventually be using the ZIP logic, so this is good to know.

Issue #294, opened 2017-07-22
Reflection support


sriharshachilakapati:
I was browsing through the sources of TeaVM as I'm currently evaluating whether I can safely migrate my existing code base (~2000 classes and interfaces) from GWT to it, and a commit caught my attention.

12dded7 - JS: add reflection support for methods

This is interesting, I browsed through the wiki and the main site teavm.org but can't find any other information on this. Is this possible? If it is, then how should I be using it?

My use case is that I'm a game developer, and I currently have a mechanism that allows me to use tools to create entities. These entities will contain the class names of the components attached to them, and are written to text files, and at runtime I instantiate the components based on the name.

~~~java
Class<? extends Component> cls = Class.forName(componentClassName);
// ... instantiate it to get a Component object and add it to the entity
~~~

Is this kind possible in TeaVM? In GWT, I'm currently using Generators to write a new class that returns the object at compile time that enumerates all the component class names in the compile path.

If the only issue for reflection is the size, then it may be a good idea to allow for a property that selects the classes that should be kept at all cost?

konsoletyper:
There will be method and field reflection support in TeaVM 0.6.0, however it's still far from being released, so I haven't provided any documentation yet. If you want to use reflection, you may build TeaVM yourself and use version from your local Maven repository.

As for current versions, reflection *is* available, but it's very limited. You can't access fields and methods of classes. These commits you mention don't introduce "full reflection support", they just add some features which weren't available before.

You can use `Class.forName` and `Class.newInstance()` from 0.5.x, with only one limitation: you have to mention classes available from `Class.forName` somewhere in the reachable methods (`main` method is a good candidate). This may be, for example, a class literal you pass to some function. Example is:

```
static void main(String[] args) {
    someMethod9(A.class, B.class);
}

static void someMethod(Class<?>... classes) {
    for (Class<?> c : classes) {
        System.out.println(c.getSimpleName());
    }
}
```

The reason is: to produce small JS files, TeaVM will only compile classes/methods it can observe, and won't compile any other classes.

[...]

PR #399, opened 2019-06-23
Added some function interfaces that I found to be missing whilst test…

WolfieWerewolf:
…ing reflection.  Note that I removed the style checker from my local POM (Which I have not comitted) so you may need to modify for style... All of the existing unit tests pass.  I updated an open ticket about reflection #294 with some additional information.

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -31,6 +31,7 @@
     <inspection_tool class="Convert2Lambda" enabled="true" level="WEAK WARNING" enabled_by_default="true" />
     <inspection_tool class="Convert2MethodRef" enabled="true" level="WEAK WARNING" enabled_by_default="true" />
     <inspection_tool class="CssUnusedSymbol" enabled="true" level="WEAK WARNING" enabled_by_default="true" />
+    <inspection_tool class="Duplicates" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
     <inspection_tool class="EmptyCatchBlock" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="m_includeComments" value="true" />
       <option name="m_ignoreTestCases" value="true" />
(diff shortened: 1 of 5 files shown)


classlib/src/main/java/org/teavm/classlib/java/util/concurrent/TConcurrentMap.java
```diff
@@ -0,0 +1,169 @@
+/*
```
konsoletyper: Please, fix checkstyle errors in this file

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -31,6 +31,7 @@
     <inspection_tool class="Convert2Lambda" enabled="true" level="WEAK WARNING" enabled_by_default="true" />
     <inspection_tool class="Convert2MethodRef" enabled="true" level="WEAK WARNING" enabled_by_default="true" />
     <inspection_tool class="CssUnusedSymbol" enabled="true" level="WEAK WARNING" enabled_by_default="true" />
+    <inspection_tool class="Duplicates" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
```
konsoletyper: Please, remove this change from the first commit. It does not relate to other changes, and I think that changing IDEA inspection is a matter of a separate discussion.

classlib/src/main/java/org/teavm/classlib/java/util/concurrent/TConcurrentMap.java
konsoletyper: BTW, where did you get the code?

.idea/inspectionProfiles/Project_Default.xml
WolfieWerewolf: Alexey, re the IDEA inspection my error, I will fix that.  I will re-enable checkstyles and fix that too.. sorry.  I got the code from openjdk8.  Is that acceptable?  Perhaps I need to include their license somewhere?

konsoletyper: You can't include any code from OpenJDK, since it's licensed under GPL, which is not compatible with Apache license.

WolfieWerewolf: Ok that makes it a bit tricky.  I will see if Harmony or GWT has these and let you know.

Ihromant:
All classes from this PR are in codebase already.

PR #410, opened 2019-08-06
ClassGenerator.prepare: handle IsInstanceInstruction case

volth:
C-generator sometimes does not emit 1-2 files because of it

### DIFF
--- a/core/src/main/java/org/teavm/backend/c/generate/ClassGenerator.java
+++ b/core/src/main/java/org/teavm/backend/c/generate/ClassGenerator.java
@@ -178,6 +178,11 @@ public class ClassGenerator {
         }
 
         @Override
+        public void visit(IsInstanceInstruction insn) {
+            addType(insn.getType());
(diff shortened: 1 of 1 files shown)


volth:
If you need a testcase, try to compile this to C:

```scala
package org.teavm.samples.scala
object Client {
  def main(args: Array[String]) {
    val bs = collection.mutable.BitSet.empty
    bs += 13
    println(s"bs=${bs}")
  }
}
```

konsoletyper:
Good! How did you found C backend and how did you learn how to use it? BTW, there are some issues in GC that I fixed in gc-defrag branch, but I can't merge it into master until I test it a little bit more. So if you find some issues in GC, please, try this branch before filing an issue.

volth:
I do not remember, I have old snippets here. Just spotted recent activity related to C backend and gave it a try.

Yes, `gc-defrag` works where `master` crashes

PR #414, opened 2019-08-07
sanitize unicode in java identifiers

volth:
greek letters and math symbols (`λ`, etc) are ok as Java identifiers C compilers do not accept them.

### DIFF
--- a/core/src/main/java/org/teavm/backend/lowlevel/generate/LowLevelNameProvider.java
+++ b/core/src/main/java/org/teavm/backend/lowlevel/generate/LowLevelNameProvider.java
@@ -191,10 +191,14 @@ public abstract class LowLevelNameProvider {
                 case '>':
                 case '<':
                 case '$':
+                case '_':
                     sb.append('_');
                     break;
                 default:
-                    sb.append(c);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/teavm/backend/lowlevel/generate/LowLevelNameProvider.java
```diff
@@ -191,10 +191,15 @@ private String sanitize(String name) {
...
                 case '$':
+                case '_':
                     sb.append('_');
                     break;
                 default:
-                    sb.append(c);
+                    if ('a' <= c && c <= 'z' || 'A' <= c && c <= 'Z' || '0' <= c && c <= '9' && i != 0) {
+                        sb.append(c);
+                    } else {
+                        sb.append(String.format("U%04X", (int) c));
```
konsoletyper: I think it's ok just to append '_' character here. Also, I usually avoid using of String.format

volth: than `λ` and `σ` both will have `_` name which result in conflict

konsoletyper: This method is expected to return same names for different methods. Generation of non-conflicting names is handled in another place.

PR #415, opened 2019-08-07
teavm_printString: fix on linux

volth:
variable `c` was never assigned, so stack dumps were printed incorrectly (only first letters of each line)

### DIFF
--- a/core/src/main/resources/org/teavm/backend/c/runtime.c
+++ b/core/src/main/resources/org/teavm/backend/c/runtime.c
@@ -296,15 +296,15 @@ void teavm_printString(char16_t* s) {
             wchar_t* buf = malloc(sizeof(wchar_t) * cap);
             wchar_t* out = buf;
             int32_t sz = 0;
-            wchar_t c;
-            do {
+            while (*s != '\0') {
                 s = teavm_utf16ToUtf32(s, out++);
                 if (++sz == cap) {
                     cap *= 2;
                     buf = realloc(buf, sizeof(wchar_t) * cap);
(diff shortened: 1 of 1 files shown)


core/src/main/resources/org/teavm/backend/c/runtime.c
```diff
@@ -296,15 +296,15 @@ void teavm_printString(char16_t* s) {
...
+            while (*s != '\0') {
                 s = teavm_utf16ToUtf32(s, out++);
                 if (++sz == cap) {
                     cap *= 2;
                     buf = realloc(buf, sizeof(wchar_t) * cap);
                     out = buf + sz;
                 }
-            } while (c != '\0');
+            }
+            *out = '\0';
```
konsoletyper: What happens if there's not enough memory allocated for extra '\0' character? The code was written this way for purpose.

konsoletyper: There's a proper fix in gc-defrag branch: https://github.com/konsoletyper/teavm/commit/b9a189093ba9c72a39ae65c03dfc67b6e7757989

volth: There is always 1 cell in the output buffer, where `teavm_utf16ToUtf32` writes.

`do {} while`-loop would just perform a needless `realloc` if `\0' fits into the last cell in the buffer.

volth: actually I put an `assert()` there to be sure :)
just removed before PR as it was not hit.

PR #425, opened 2019-09-30
Minor improvments for javascript runtime

MaxGraey:


konsoletyper:
Ok, I fixed problem in JavaScript parser and now tests should pass normally. Please, rebase to latest master and re-run tests. Not all of them pass. For example `StringBuilderTest.nanFloatAppended`. Looks like your changes affected comparison with NaN values.

MaxGraey:
Both version not handle NaN comparision. Example:
```ts
const $rt_compare = (a, b) => a > b ? 1 : a < b ? -1 : a === b ? 0 : 1;
console.log([1, NaN, Infinity, -2, 0].sort($rt_compare))
```
Output:
```
[ 1, NaN, -2, 0, Infinity ]
```
```ts
const $rt_compare = (a, b) => (a > b) - (a < b);
console.log([1, NaN, Infinity, -2, 0].sort($rt_compare))
```
Same output:
```
[ 1, NaN, -2, 0, Infinity ]
```

To properly handle NaNs you should use something like this:
```ts
function $rt_compare(a, b) {
  var nanA = isNaN(a);
  var nanB = isNaN(b);
  return nanA && !nanB ? +1 : !nanA && nanB ? -1 : (a > b) - (a < b);
}

console.log([1, NaN, Infinity, -2, 0].sort($rt_compare));
```
Output:
```
[ -2, 0, 1, Infinity, NaN ]
```

konsoletyper:
Why don't you just rollback to former implementation of `$rt_compare`?

MaxGraey:
> Why don't you just rollback to former implementation of $rt_compare?

Because it not cover all `NaN` cases and not handle `+0` and `-0`. See tail part of [comment below](https://github.com/konsoletyper/teavm/pull/425#issuecomment-536640509)

MaxGraey:
Previous:
```js
[0, -0, -1, NaN, Infinity, 2].sort((a, b) => a > b ? 1 : a < b ? -1 : a === b ? 0 : 1);
> [-1, 0, -0, NaN, 2, Infinity]
```
New:
```js
function compare(a, b) {
    var nanA = isNaN(a);
    var nanB = isNaN(b);
    if (nanA && !nanB) return 1;
    if (!nanA && nanB) return -1;
    if (a === 0 && b === 0) {
      a = 1 / a;
      b = 1 / b;
    }
    return (a > b) - (a < b);
}
[0, -0, -1, NaN, Infinity, 2].sort(compare);

> [-1, -0, 0, 2, Infinity, NaN]
```

konsoletyper:
Sorry, I like you entire commit except for `$rt_compare`. The idea was to make code shorter and faster. Your implementation of `$rt_compare` is shorter and faster, but less accurate. Now you propose more verbose and more slow version of `$rt_compare` (and more accurate), which does not fit into whole idea of this commit. I don't think `$rt_compare` should slow. Yes, there's problem with accuracy of `$rt_compare`, but this problem should be solved with performance in mind (and should not be solved just in runtime level, but deep in compiler level). So please just roll back this function and I approve your PR, and if you want more precise behaviour of `$rt_compare`, let's discuss this in a separate place.

MaxGraey:
Why you think this much slower? I slightly improve this version and that's bench results:
Chrome 77:
crm77

Firefox 69:

FF69


[Benchmark suite](https://esbench.com/bench/5d9251b84cd7e6009ef627ef)

So it's just ~5% overhead

MaxGraey:
But if you want I'll roll back this

MaxGraey:
Done!

MaxGraey:
But you should know this will be different to [Java's behaviour](https://code.sololearn.com/cVRUy2BwauK8)

MaxGraey:
I updated [bench](https://esbench.com/bench/5d9251b84cd7e6009ef627ef) to this version:
```ts
function enchancedComparator(a, b) {
    var nanA = a != a;
    var nanB = b != b;
    if (nanA | nanB) return nanA - nanB;
    if (a === 0 && b === 0) { a = 1 / a, b = 1 / b; }
    return (a > b) - (a < b);
}
```

And now results even better:
new-res

konsoletyper:
Since your new version is faster and more precise then the old one, I think you can publish it in this PR. But please make sure that all tests pass

PR #429, opened 2019-10-06
Fill in some gaps in the class library

colorizenl:
- the inner classes are marked as public in the JDK, but not in TeaVM's version
- added Color/Dimension/Point, these are defined in AWT, but also accessed from elsewhere in the JDK
- added UUID emulation
- added more logging classes so that applications don't immediately crash when trying to initialize logging


konsoletyper:
Please, fix code style issues reported by checkstyle

colorizenl:
Done!

PR #432, opened 2019-10-10
#431: Add support for char annotation methods

### LINKED ISSUE #431: Parser.parseAnnotationValue fails on annotation methods returning char
When an annotation has a method returning `char`, TeaVM 0.6.0 throws `AssertionError`.

fluorumlabs:
Fixes #431

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/reflect/AnnotationDependencyListener.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/reflect/AnnotationDependencyListener.java
@@ -257,6 +257,8 @@ public class AnnotationDependencyListener extends AbstractDependencyListener {
         switch (value.getType()) {
             case AnnotationValue.BOOLEAN:
                 return pe.constant(value.getBoolean() ? 1 : 0);
+            case AnnotationValue.CHAR:
+                return pe.constant(value.getChar());
             case AnnotationValue.BYTE:
                 return pe.constant(value.getByte());
             case AnnotationValue.SHORT:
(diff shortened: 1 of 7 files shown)


fluorumlabs:
I modified `MetaprogrammingTest.annotationsWork` test, but, for some reason it was passing even for failing assertion. I was running it with `mvn test -Dtest=MetaprogrammingTest`, and in generated `js` files I see the correct expected/actual values. No idea if it's my setup failing, or something else is wrong.

konsoletyper:
I'm afraid this commit is incomplete. You should support these methods in `AnnotationIO` and `ModelUtils.copyAnnotationValue`.

fluorumlabs:
I knew I missed something :) Will do.

fluorumlabs:
Could you please elaborate a bit? I added support for `CHAR` both to `AnnotationIO.writeAnnotationValue` and `AnnotationIO.readAnnotationValue`. `ModelUtil.copyAnnotationValue` however, does not care about primitive types:

```java
    private static AnnotationValue copyAnnotationValue(AnnotationValue value) {
        switch (value.getType()) {
            case AnnotationValue.LIST: {
                List<AnnotationValue> listCopy = new ArrayList<>();
                for (AnnotationValue item : value.getList()) {
                    listCopy.add(copyAnnotationValue(item));
                }
                return new AnnotationValue(listCopy);
            }
            case AnnotationValue.ANNOTATION:
                return new AnnotationValue(copyAnnotation(value.getAnnotation()));
            default:
                return value;
        }
    }
```

konsoletyper:
Ah, you right. You don't need to copy primitive annotation values

PR #441, opened 2019-11-17
Support for System.getenv()

colorizenl:
This will always return that the environment variables does not exist, it is purely intended for compatibility with applications that check for optional environment variables (in those cases the TeaVM transpiler currently fails because the method does not exist).


konsoletyper:
I think it's reasonable to provide fair implementation for this feature for C backend.

PR #443, opened 2019-11-24
Extend the class library with the part of Spliterators that is used by Google Guava

colorizenl:
The class library already contained some support for the Spliterator itself, but Guava also uses some methods from the Spliterators utility class.


classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.function.IntConsumer;
+
+public class TSpliterators {
+
+    private TSpliterators() {
+    }
+
+    public static <T> TSpliterator<T> spliterator(Object[] array, int additionalCharacteristics) {
```
konsoletyper: Why didn't you implement Spliterator for this case?

classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
```diff
@@ -0,0 +1,77 @@
...
+            int index = fromIndex;
+
+            @Override
+            public TSpliterator.OfInt trySplit() {
+                return null;
+            }
+
+            @Override
+            public boolean tryAdvance(IntConsumer action) {
+                index++;
```
konsoletyper: Does this work properly. Can you add tests for these methods?

classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
```diff
@@ -0,0 +1,77 @@
...
+                } else {
+                    return false;
+                }
+            }
+        };
+    }
+
+    public static <T> TSpliterator<T> spliterator(Iterator<? extends T> iterator,
+                                                  long size, int characteristics) {
+        TArrayList<T> list = new TArrayList<T>();
```
konsoletyper: Why didn't you provide fair implementation for this kind of spliterator?

classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
colorizenl: Not sure I understand. The "normal" Spliterators also doesn't implement Spliterator. Could it be that you misread TSpliterator instead of TSpliterators? If not, I'm probably misunderstanting something.

classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
colorizenl: Added unit tests. And you were indeed correct they were needed because the initial version contained a mistake.

classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
colorizenl: Changed the implementation to be based on the skeleton provided by TSpliterator.

classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
konsoletyper: You create extra list here, while you could just implement Spliterator interface that exposes array content directly. This would (1) consume less memory (2) reflect changes to underlying array. For the pt.2 you should write tests to ensure that your implementation works as in OpenJDK

classlib/src/test/java/org/teavm/classlib/java/util/TSpliteratorsTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.IntConsumer;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class TSpliteratorsTest {
```
konsoletyper: Please note how other test clases named. You should omit `T` prefix here

classlib/src/test/java/org/teavm/classlib/java/util/TSpliteratorsTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    public void testIntSpliterator() {
```
konsoletyper: You should omit `test` prefix here.

classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
colorizenl: Changed implementation and added test for the behavior suggested in item 2.

classlib/src/main/java/org/teavm/classlib/java/util/TSpliterators.java
```diff
@@ -0,0 +1,120 @@
...
+    public static <T> TSpliterator<T> spliterator(Collection<? extends T> c, int characteristics) {
+        return ((TCollection) c).spliterator();
+    }
+
+    public static TSpliterator.OfInt spliterator(int[] array, int additionalCharacteristics) {
+        return spliterator(array, 0, array.length, additionalCharacteristics);
+    }
+
+    public static TSpliterator.OfInt spliterator(int[] array, int fromIndex, int toIndex,
+                                                 int additionalCharacteristics) {
```
konsoletyper: FYI: TeaVM code base does not follow this convention with indentation of parameters on single column, use 8 spaces instead.

konsoletyper:
Another thing I noticed after I merged PR: you placed SpliteratorsTest class in a wrong module

colorizenl:
@konsoletyper sorry about that. I just used the default of src/test/java, didn't realize until now there was a separate tests project. Will remember it for the next time.

PR #444, opened 2019-11-28
Gamepad api

devnewton:
Add https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API to JSO.

Usefull for video games like mine, Newton Adventure, that I'm currently porting from to Teavm.

screenshot

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/dom/events/GamepadEventTarget.java
+++ b/jso/apis/src/main/java/org/teavm/jso/dom/events/GamepadEventTarget.java
@@ -1,0 +1,37 @@
+/*
+ *  Copyright 2019 devnewton.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


jso/apis/src/main/java/org/teavm/jso/dom/events/GamepadEventTarget.java
```diff
@@ -0,0 +1,37 @@
...
+
+    default void neglectGamepadConnected(EventListener<GamepadEvent> listener) {
+        removeEventListener("gamepadconnected", listener);
+    }
+
+    default void listenGamepadDisconnected(EventListener<GamepadEvent> listener) {
+        addEventListener("gamepaddisconnected", listener);
+    }
+
+    default void neglectlistenGamepadDisconnected(EventListener<GamepadEvent> listener) {
```
konsoletyper: Please, fix this name

devnewton: done

PR #446, opened 2020-01-05
Add support for URL encode/decode methods added in Java 10

colorizenl:
Java 10 added versions of URL encode/decode that take the Charset instance as an argument rather than the charset name. This means the user of these methods no longer has to handle the UnsupportedEncodingException (since the presence of a Charset instance indicates the character encoding is known to exist).

This method was already present in the URLDecoder, but not yet in the URLEncoder.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/net/TURLEncoder.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/net/TURLEncoder.java
@@ -53,24 +53,8 @@ public final class TURLEncoder {
     }
     
     public static String encode(String s, Charset enc) {
-        try {
-            return encode(s, enc.displayName());
-        } catch (UnsupportedEncodingException e) {
-            // Cannot happen as the provided character encoding is known to be available.
-            throw new AssertionError(e);
-        }
-    }
-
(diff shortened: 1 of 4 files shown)


classlib/src/main/java/org/teavm/classlib/java/net/TURLEncoder.java
```diff
@@ -51,6 +51,15 @@ public static String encode(String s) {
         }
         return buf.toString();
     }
+    
+    public static String encode(String s, Charset enc) {
```
konsoletyper: This approach to implement this method is wrong. The primary logic should be in this method, while the former one (that which takes charset name) should delegate to this method.

colorizenl: I see your point, done. Note that I could not add a new test in `URLEncoderTest.java`, as this would make the test itself incompatible with Jaba 8 (which is currently the baseline for this project). However, the changed code is still tested via the existing test cases for `encode(String, String)`.

konsoletyper:
Sorry, I forgot to merge this PR. However, you added commit that is not related to initial purpose of PR. Please, remove it so that I can squash commits related to URL

PR #453, opened 2020-01-11
Add write replace method to lambdas that implement java.io.Serializable

adamjryan:
* Peripheral java library classes added as part of the toString method
of SerializedLambda to align functionality with the standard library.
* Test added that checks the SerializedLambda can be returned and that
it is populated with the correct values including it's toString method.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/invoke/TSerializedLambda.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/invoke/TSerializedLambda.java
@@ -1,0 +1,114 @@
+/*
+ *  Copyright 2020 adam.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.classlib.java.lang.invoke;
+
+import java.io.Serializable;
+import java.lang.invoke.MethodHandleInfo;
+import java.util.Objects;
+
+public final class TSerializedLambda implements Serializable {
+    private static final long serialVersionUID = 8025925345765570181L;
+    private final Class<?> capturingClass;
+    private final String functionalInterfaceClass;
+    private final String functionalInterfaceMethodName;
+    private final String functionalInterfaceMethodSignature;
+    private final String implClass;
+    private final String implMethodName;
+    private final String implMethodSignature;
+    private final int implMethodKind;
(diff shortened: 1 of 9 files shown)


konsoletyper:
I have no idea why you need this. Serialization is not supported in TeaVM and will never be.

adamjryan:
I'm not sure I follow. It is not the intention of this change to mandate that TeaVm support the JDK serialisation mechanisms. The writeReplace method is the only way to reflect on the metadata of a lambda, which is something I have a use for.

konsoletyper:
> The writeReplace method is the only way to reflect on the metadata of a lambda, which is something I have a use for.

What use?

classlib/src/main/java/org/teavm/classlib/java/lang/invoke/TMethodHandleStatics.java
```diff
@@ -0,0 +1,96 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.classlib.java.lang.invoke;
+
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+class TMethodHandleStatics {
```
konsoletyper: This class looks like copy of OpenJDK class, so I can't accept since it violates OpenJDK license

classlib/src/main/java/org/teavm/classlib/java/lang/invoke/TSerializedLambda.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private Object readResolve() throws ReflectiveOperationException {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String toString() {
+        String implKind = MethodHandleInfo.referenceKindToString(implMethodKind);
+        return String.format("SerializedLambda[%s=%s, %s=%s.%s:%s, "
```
konsoletyper: You better prefer string concatenation here

classlib/src/main/java/org/teavm/classlib/java/lang/invoke/TMethodHandleStatics.java
adamjryan: It's not OpenJDK. I've removed anyway.

adamjryan:
> > The writeReplace method is the only way to reflect on the metadata of a lambda, which is something I have a use for.
> 
> What use?

I'm using them as method pointers against an interface for a remote server hosted object.

core/src/main/java/org/teavm/model/MethodHandleType.java
```diff
@@ -16,6 +16,7 @@
 package org.teavm.model;
 
 public enum MethodHandleType {
+    REF_NONE,
```
konsoletyper: For what purpose? How are you going to use it? Found no references to this field. Also, adding such things in the core model requires **a lot** of things to be changed, like LambdaMetatafactorySubstitutor, model serializer and deserializer, etc, etc.

adamjryan: This was to align with the standard Java library. I've added a referenceKind value instead now on the enum so it will be backwards compatible.

Issue #457, opened 2020-01-12
Consider supporting setCustomValidity


hohwille:
The w3c specifies [HTMLObjectElement.setCustomValidity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/setCustomValidity).
I would suggest that in TeaVM `jso-apis` an `HTMLObjectElement` will be introduced with the method `setCustomValidity(String customValidity)` and `HTMLInputElement` should extend this new  interface`HTMLObjectElement`.
Any concerns? Should I provide a PR?

konsoletyper:
To fix issues like these you only need trivial change, so it's better to provide PR right away.

hohwille:
Sure, will do that next time omitting the issue. I was not aware how you deal with these things as every github project does this differently. Further, I was not aware if the `HTML*Element` types are missing out some properties on purpose (e.g. for ultimate performance). However, my understanding of TeaVM is that code that is not used, should also not affect run-time-performance (but maybe minimal impact on size and compile-time performance).

hohwille:
As PR #458 was merged, this issue can now be closed.

PR #458, opened 2020-01-13
#457: added setCustomValidity

hohwille:
fix for #457


jso/apis/src/main/java/org/teavm/jso/dom/html/HTMLInputElement.java
```diff
@@ -65,6 +65,9 @@
 
     @JSProperty
     void setValue(String value);
+    
+    @JSProperty
```
konsoletyper: Please, remove `@JSProperty` annotations from `setCustomValidity` methods, since they are methods, not properties.

hohwille: Thanks for the feedback. Got it. Will update immediately.

Issue #445, opened 2019-12-18
Add missing java.util.concurrent.ConcurrentHashMap


niloc132:
Looking at the [LinkedBlockingDeque](https://github.com/konsoletyper/teavm/blob/master/classlib/src/main/java/org/teavm/classlib/java/util/concurrent/TLinkedBlockingDeque.java) implementation, it appears that this could be done the same way that it is in GWT - simply wrap HashMap and implement ConcurrentMap and add those missing methods. 

However, this might not be fully correct, given the coroutine implementation - `@Async` JS code could potentially be used for equals/hashcode, or some lambda could call async code or manipulate the contents of the map while running, etc.

hohwille:
See also #449 that I filed today.

hohwille:
@niloc132 
> Looking at the LinkedBlockingDeque implementation, it appears that this could be done the same way that it is in GWT - simply wrap HashMap and implement ConcurrentMap and add those missing methods.

I created PR #462 for this but seems only to be acceptable if the implementation is feature complete. You can take the code from my PR and copy it to your project though as a workaround.

PR #462, opened 2020-01-22
#445: Added support for Concurrent[Hash]Map

hohwille:
Simplest fix for #445.
Actually in recent JDK (9+) `ConcurrentMap` only omtimizes some default methods of `Map` but does not add any specific methods anymore since they have all been added to `Map` already (IMHO even in JDK8) whereas in JDK7 `ConcurrentMap` was having lots of extra methods.
As even `TMap` already contains all these methods (e.g. `getOrDefault`) there is even no issue with older JDKs.
As inside the browser cuncurrency is not an issue like in JVM the simplest approach is to let `ConcurrentHashMap` behave exactly like `HashMap`.
Therefore I added this minimalistic approach that I tested with OpenJDK12 and recent TeaVM with my patch. I also did not have to "copy" any code that might be copyrighted by Oracle or so for this minimal solution.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/concurrent/TConcurrentHashMap.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/concurrent/TConcurrentHashMap.java
@@ -1,13 +1,35 @@
+/*
+ *  Copyright 2020 Joerg Hohwiller.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
 package org.teavm.classlib.java.util.concurrent;
 
 import org.teavm.classlib.java.util.THashMap;
 import org.teavm.classlib.java.util.TMap;
 
+// currently behaves like regular map, concurrency has to be added
+// https://github.com/konsoletyper/teavm/issues/445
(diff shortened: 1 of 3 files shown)


konsoletyper:
> As inside the browser cuncurrency is not an issue like in JVM the simplest approach is to let ConcurrentHashMap behave exactly like HashMap.

Wrong, concurrency matters, since TeaVM emulates it, so implementing ConcurrentHashMap is going to be non-trivial task.

hohwille:
> Wrong, concurrency matters, since TeaVM emulates it, so implementing ConcurrentHashMap is going to be non-trivial task.

Thanks for your feedback and clarification. OK. However, currently when I use Concurrent[Hash]Map in my code, the TeaVM compiler fails. With this PR TeaVM would pass and the map behaves like a map even though it might not satisfy its contract. I never considered using creating Threads in my client code. 

So can we consider agile improvements step by step and get this merged and create another issue to properly implement TConcurrentHashMap? I surely will quickly add a header to make checkstyle happy here.

konsoletyper:
> OK. However, currently when I use Concurrent[Hash]Map in my code, the TeaVM compiler fails

You can workaround it by including `TConcurrentHashMap` to your code base.

> I never considered using creating Threads in my client code

But there are people who use thread in their client code.

> So can we consider agile improvements step by step and get this merged and create another issue to properly implement TConcurrentHashMap? I surely will quickly add a header to make checkstyle happy here.

No, I can't approve this. I think ConcurrentHashMap implementation should be complete. I tried to do that, but I failed to implement it in the way it's efficient in single-threaded code. Things are quite complicated there.

hohwille:
Please note that I have to replace usages of `ConcurrentHashMap` with `HashMap` for `java.time` support in #467 so your decision is not really helping for accuracy on the long run.

konsoletyper:
Can you point on a certain places where you replaced `ConcurrentHashMap` with `HashMap`?

hohwille:
Search for `ConcurrentHashMap` in this commit:
https://github.com/konsoletyper/teavm/pull/467/commits/0724f154caec10dc00fe47c94cd0df148f1a3229

icecreamparlor:
Any updates ..?

konsoletyper:
@icecreamparlor no, I won't approve this PR. Instead, I'll implement ConcurrentHashMap myself some day.

Issue #451, opened 2020-01-10
Support for java.time


hohwille:
TeaVM does not seem to support `java.time` (JSR310).
I get such errors:
```
[ERROR] Class java.time.Instant was not found
[ERROR] Class java.time.LocalDate was not found
[ERROR] Class java.time.LocalDateTime was not found
[ERROR] Class java.time.LocalTime was not found
[ERROR] Class java.time.OffsetDateTime was not found
[ERROR] Class java.time.OffsetTime was not found
[ERROR] Class java.time.ZonedDateTime was not found
[ERROR] Class java.time.LocalDateTime was not found
[ERROR] Class java.time.OffsetDateTime was not found
[ERROR] Class java.time.LocalDate was not found
[ERROR] Class java.time.OffsetTime was not found
[ERROR] Class java.time.LocalTime was not found
[ERROR] Class java.time.Instant was not found
[ERROR] Class java.time.ZonedDateTime was not found
```

hohwille:
That would also be some area where I might assist. 
Looking at [TDate](https://github.com/konsoletyper/teavm/blob/master/classlib/src/main/java/org/teavm/classlib/java/util/TDate.java) this does not look like an easy task and `java.time` is huge and complex.

hohwille:
I started working on this. Here are some general thoughts:
* support for non ISO chonology is IMHO overkill for a browser hence Japanese, Hijrah, Minguo, and ThaiBuddhist will not be supported.
* `java.time` has advanced serialization build in (`writeReplace`, `readObject`, etc.). As standard java serialization does IMHO not make any sense in the browser this should also not be supported.
* `ZoneId` should be supported to some degree. At least `UTC` has to work so it makes any sense. However, also `ZoneOffset` would be nice to have. But region IDs would most probably be overkill again. They could be supported as containing any kind of string but not allowing to calculate with it as the offsets for all regions require a larger map of data.
* In general I would prefer an approach with compatibility rather than maximum performance.
* Tricky aspect is the `DateTimeFormatter[Builder]`
* In question is if `TemporalQuery` should actually be supported.

hohwille:
When I write `T*` types for `classlib` I can use the regular java type as well as the `T*` type. From what I have seen in `classlib` I assume that using the `T*` types wherever possible is preferred. In cases like `toString()` that can not return `TString` the plain Java type is used.
[...]

PR #464, opened 2020-01-24
#451: added java.time support

hohwille:
WIP for #451

TODOs:
* [ ] TMath needs to be extended to support required calculation methods (at least as dummy emulation).
* [ ] I disabled checkstyle as it is way too pick. If that is seriously a requirement, someone else need to help. Further I will and can not add copyright headers stating that the code is written by Alexey Andreev. However, I have an old Email from jodastephen that I can copy his code from https://github.com/ThreeTen/threetenbp
* [ ] for the design I am thinking about deleting the entire format package, remove support for custom format/parse methods and hardcode ISO parsing and formatting to keep the performance and size fine for the browser. WDYT?

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/time/TClock.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/time/TClock.java
@@ -1,0 +1,360 @@
+package org.teavm.classlib.java.time;
+
+import static org.teavm.classlib.java.time.TLocalTime.NANOS_PER_MILLI;
+import static org.teavm.classlib.java.time.TLocalTime.NANOS_PER_MINUTE;
+import static org.teavm.classlib.java.time.TLocalTime.NANOS_PER_SECOND;
+
+import org.teavm.classlib.java.io.TSerializable;
+import org.teavm.classlib.java.lang.TIllegalArgumentException;
+import org.teavm.classlib.java.lang.TSystem;
+import org.teavm.classlib.java.util.TObjects;
+
+public abstract class TClock {
+
+  public static TClock systemUTC() {
+
+    return TSystemClock.UTC;
+  }
+
+  public static TClock systemDefaultZone() {
+
+    return new TSystemClock(TZoneId.systemDefault());
+  }
+
+  public static TClock system(TZoneId zone) {
+
+    TObjects.requireNonNull(zone, "zone");
+    if (zone == TZoneOffset.UTC) {
+      return TSystemClock.UTC;
+    }
+    return new TSystemClock(zone);
+  }
+
+  public static TClock tickMillis(TZoneId zone) {
+
+    return new TTickClock(system(zone), NANOS_PER_MILLI);
+  }
+
+  public static TClock tickSeconds(TZoneId zone) {
+
+    return new TTickClock(system(zone), NANOS_PER_SECOND);
+  }
+
+  public static TClock tickMinutes(TZoneId zone) {
+
+    return new TTickClock(system(zone), NANOS_PER_MINUTE);
+  }
+
+  public static TClock tick(TClock baseClock, TDuration tickDuration) {
+
+    TObjects.requireNonNull(baseClock, "baseClock");
+    TObjects.requireNonNull(tickDuration, "tickDuration");
+    if (tickDuration.isNegative()) {
+      throw new TIllegalArgumentException("Tick duration must not be negative");
+    }
+    long tickNanos = tickDuration.toNanos();
+    if (tickNanos % 1000_000 == 0) {
+    } else if (1000_000_000 % tickNanos == 0) {
+    } else {
+      throw new TIllegalArgumentException("Invalid tick duration");
+    }
+    if (tickNanos <= 1) {
+      return baseClock;
+    }
+    return new TTickClock(baseClock, tickNanos);
+  }
+
+  public static TClock fixed(TInstant fixedInstant, TZoneId zone) {
+
+    TObjects.requireNonNull(fixedInstant, "fixedInstant");
+    TObjects.requireNonNull(zone, "zone");
+    return new TFixedClock(fixedInstant, zone);
+  }
+
+  public static TClock offset(TClock baseClock, TDuration offsetDuration) {
+
+    TObjects.requireNonNull(baseClock, "baseClock");
+    TObjects.requireNonNull(offsetDuration, "offsetDuration");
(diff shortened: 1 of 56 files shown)


konsoletyper:
Where did you get this code? Is this from OpenJDK? If it's from OpenJDK, I can't approve this PR, since it violates GPL license

hohwille:
In https://github.com/ThreeTen/threetenbp License is BSD (https://github.com/ThreeTen/threetenbp/blob/master/LICENSE.txt). As I said I have connection to the author @jodastephen and we can ask him for official confirmation of using his code in TeaVM.

konsoletyper:
> Further I will and can not add copyright headers stating that the code is written by Alexey Andreev

You can add copyright headers stating that code was written by Jörg Hohwiller. It's a derivative work based on original code written by threeten contributors. This can be stated in NOTICE file.

konsoletyper:
> for the design I am thinking about deleting the entire format package, remove support for custom format/parse methods and hardcode ISO parsing and formatting to keep the performance and size fine for the browser

No, please, don't. Current java.Date formatting is feature complete. This is kind of trade-off, but my vision is that if something's implemented in TeaVM, it should work as much Java as possible. Also, if you really care about performance and code size you can try to add a TeaVM option that enables reduced support for java.time formatting API.

konsoletyper:
What about time zones? Did you incorporate time zone support from java.util.TimeZone implementation?

konsoletyper:
What about tests?

jodastephen:
For the record, TheeTen-Backport project is [BSD licensed](https://github.com/ThreeTen/threetenbp/blob/master/LICENSE.txt) and can be used in accordance with that license.

hohwille:
@jodastephen thanks for your confirmation about the BSD licensing. TeaVM is ASL 2.0 though:
https://github.com/konsoletyper/teavm/blob/master/LICENSE
This PR will still have a long way to go and code might change a lot. We might then come back to you if we need a final agreement to license code under ASL 2.0. I might even recreate the entire code-base for threeten backport with some automated scripts (and also check the diffs to current state and also to OpenJDK, etc.).

hohwille:
> What about tests?

The treeten-bp project also has lots of tests. I could try to also migrate those to test the TeaVM adopted code. If I remeber correctly they are based on TestNG.

hohwille:
> What about time zones? Did you incorporate time zone support from java.util.TimeZone implementation?

No. Way too complex and depending on complex JDK internals we can not easily emulate.
I planned to only support ZoneOffset and simple things like UTC, GMT, GMT0.
Nothing like Europe/Berlin, etc.
After all, I would be very happy if we have this PR in a state that is working at all, with legal issues clarified and tests running. This already will be a lot of invest to get there still.
So actually I wanted to start this very agile and just introduce first types like Month and Year in small PRs that can be merged step by step. However, `java.time` is a deeply connected web where everything depends on everything else. So I quickly dropped this plan.
However, I can tell you that stuff like JapaneseChronology will be overkill and IMHO already the Zone-DB will kill the bootstrapping of a fast browser app. So let us start to get started with a working subset and then further PRs could actually add further support (as long as it is guaranteed that either performance still is suitable or some compiler-option/feature-toggle/... is available to choose between performance and higher compatibility).

hohwille:
> No, please, don't. Current java.Date formatting is feature complete. 

OK. Thanks for your feedback and making this clear. However have a look and see that it is much more complex than `java.util.DateFormatter` and it requires creating many builders that create a default ISO formatter with all features and options of the generic formatter just to boostrap a single `Instant`. Apart from that an `Instant` is just a container for two long values derived from `Clock` that delegates to `System` features while so far I ignored the nano adjustment as this is not available in the browser. Requires deep VM internal features. I added some commented out code in Clock that would use `System.nanoTime()`. I would be happy if we could come to the point where we can test such things and evaluate the performance and feature impact of such decisions.

> This is kind of trade-off, but my vision is that if something's implemented in TeaVM, it should work as much Java as possible. 

Sure. But looking at average code by fundamentally supporting, Year, Month, MonthDay, Instant, LocalDate, LocalTime, LocalDateTime, OffsetDateTime, OffsetTime with default ISO parsing and formatting, we could probably make 70% of code using `java.time` work. At least this would be the most important part for getting stuff working in the browser. If we want to have 100% emulation of JDK in the browser we can see what other projects have created and how useless their results are due to performance. See here and click on one of the demos that run a JavaFx app in the browser:
https://netopyr.com/2014/12/19/javafx-in-the-browser-2/
Do not forget to grab a cup of coffee or tea while waiting till the browser loaded a JavaFx app in JS that renders 5 moving circles. I hope this is not why you called this project TeaVM... I love your work because it leads to fast code and I am impressed. That is why I am supporting your project in my spare time while actually I want to do my project and vision but have it based on TeaVM that is not yet fully ready to make it come true but very close to that point.

> Also, if you really care about performance and code size you can try to add a TeaVM option that enables reduced support for java.time formatting API.

Yes, that is what I would suggest for `ZoneRegion` and might consider for the formatter stuff.
However so far I have no clue how to implement such option or toggle with TeaVM.

hohwille:
> You can add copyright headers stating that code was written by Jörg Hohwiller. It's a derivative work based on original code written by threeten contributors. This can be stated in NOTICE file.

I would rather add the original author which is `Stephen Colebourne` for most of the classes if you really need a individual person and if required add myself as co-author. In Germany we have a specific law for authorship that is not really copyright compatible and this law prevents you from giving away or taking over the authorship of work.

konsoletyper:
> No. Way too complex and depending on complex JDK internals we can not easily emulate.
I planned to only support ZoneOffset and simple things like UTC, GMT, GMT0.
Nothing like Europe/Berlin, etc.

You don't need to depend on complex JDK internals. Timezones are already implemented in TeaVM, just reuse them! 

> Zone-DB will kill the bootstrapping of a fast browser app

You mean tzdata? Please, try java.util.TimeZone in TeaVM and see how big is the overhead.

> Do not forget to grab a cup of coffee or tea while waiting till the browser loaded a JavaFx app in JS that renders 5 moving circles.

First, even with 'heavyweight' dependencies like timezones, locales and regular expressions TeaVM shows quite good result with fast and relatively small JS. What purpose of limited support of java.time? Someone ports their application and encounter issues related to limited support of java.time, so they have to patch their code for TeaVM. IMO, they better to start getting rid of java.time API from their app at all.

hohwille:
Replaced with PR #467.

PR #478, opened 2020-02-24
Java 9 collection factories

kris-scheibe:
This adds the collection factories List.of(...), Set.of(...), Map.of(...) and Map.ofEntries(...) which were introduced to the JDK in Java 9.
It also adds the methods Map.remove(K, V) and Map.forEach(BiConsumer<K, V>) which were introduced in Java 8.

Unlike the implementation in OpenJDK, this doesn't use special implementations for the returned unmodifiable collections, but wraps the regular collection classes in an unmodifiable wrapper.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2014 Alexey Andreev.
+ *  Copyright 2020 Alexey Andreev.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
@@ -15,9 +15,8 @@
  */
 package org.teavm.classlib.java.util;
 
-import static org.teavm.classlib.java.util.TObjects.requireNonNull;
-import org.teavm.classlib.java.lang.TIllegalArgumentException;
-import org.teavm.classlib.java.lang.TNullPointerException;
+import static java.util.Objects.requireNonNull;
+import java.util.RandomAccess;
 
 /**
  * Factory-methods for List/Set/Map.of(...).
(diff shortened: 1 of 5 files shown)


classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @throws TNullPointerException if any element is null
+     */
+    @SafeVarargs
+    static <E> TList<E> createList(E... elements) {
+        if (elements == null || elements.length == 0) {
+            return TCollections.emptyList();
+        }
+
+        final TList<E> list = new TArrayList<>();
```
konsoletyper: I don't like this approach:

1. Two objects created instead of one (i.e. ArrayList and UnmodifiableList)
2. You always create a temporary array for passing varargs, even if original List.of does not use varargs
3. You rely on ArrayList which can allocate extra space in order to improve `add` performance, which is not required here in immutable list.

classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    @SafeVarargs
+    static <E> TSet<E> createSet(E... elements) {
+        if (elements == null || elements.length == 0) {
+            return TCollections.emptySet();
+        }
+
+        final TSet<E> set = new THashSet<>();
+        for (E element : elements) {
+            if (!set.add(requireNonNull(element, "element"))) {
```
konsoletyper: Is such behaviour documented for Set.of? What about tests which test this behaviour? What about tests at all?

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
```diff
@@ -15,8 +15,11 @@
  */
 package org.teavm.classlib.java.util;
 
+import static org.teavm.classlib.java.util.TObjects.requireNonNull;
+import java.util.Objects;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import org.teavm.classlib.java.util.function.TBiConsumer;
```
konsoletyper: Existing code imports from `java.util.function`. Why do you import from `org.teavm.classlib.java.util.function`?

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
```diff
@@ -15,8 +15,11 @@
  */
 package org.teavm.classlib.java.util;
 
+import static org.teavm.classlib.java.util.TObjects.requireNonNull;
```
konsoletyper: java.util.Objects.requireNotNull

classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Factory-methods for List/Set/Map.of(...).
+ */
+class CollectionsFactory {
+
+    private CollectionsFactory() {
+    }
+
+    /**
```
konsoletyper: Is there real need in such obvious doc comments? Is there any purpose on doc comments on non-public code?

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
kris-scheibe: Done.

Is there a good rule on when to use the regular java.* classes and when to use the T-classes in org.teavm.classlib.* ?
I've seen different versions of the same class used in different places, e.g.:
`static class ArrayAsList<T> extends TAbstractList<T> implements RandomAccess` in TArray vs. `public class TArrayList<E> extends TAbstractList<E> implements TCloneable, TSerializable, TRandomAccess`.
Why does one of the classes implement `TRandomAcces` and the other `RandomAccess`?

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
kris-scheibe: Done.

classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
kris-scheibe: I write a lot of library code in our company -- I guess I'm just used to document everything... but I can remove these.

classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
kris-scheibe: I wasn't sure if I should introduce new classes for these factory methods.

Do you think it's better to also have specialized implementations for the Map and Set?
THashMap and by extension THashSet aren't exactly light-weight either...

classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
kris-scheibe: Yes, all the factory methods reject null values, Set.of doesn't permit duplicate values and Map.of/ofEntries doesn't permit duplicate keys. (The classes are also supposed to be serializable, but I guess, this doesn't apply to TeaVM.)

https://docs.oracle.com/javase/9/docs/api/java/util/List.html#immutable
https://docs.oracle.com/javase/9/docs/api/java/util/Set.html#immutable
https://docs.oracle.com/javase/9/docs/api/java/util/Map.html#immutable

I can add tests to verify this behaviour.

classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
konsoletyper: I think that using HashMap and HashSet is ok, since their implementation is not so trivial.

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
konsoletyper: There's a rule to always use java package when possible. Reference to `T` class is a quite old approach and I don't use it anymore.

classlib/src/main/java/org/teavm/classlib/java/util/CollectionsFactory.java
kris-scheibe: I've added a special implementation for the immutable list, which reuses the varargs-array, if possible.
Also, I think an immutable Map/Set replacement would be a lot simpler than the HashSet implementation and could be made a bit more lightweight. If you're OK with that I'd like to work on that too, but in a new PR.

kris-scheibe:
@konsoletyper Hi, the Travis build failed while installing the JDK, can you trigger it again?

PR #482, opened 2020-03-01
added support for String.isBlank()

hohwille:
In Java11 `String.isBlank()` was added. I added support assuming it does not hurt on JDK < 11 (JDK8). Fundamental types such as `String` should IMHO be supported as widely as possible as it may confuse or frustrate users of TeaVM if their code works in Java but does not compile with TeaVM.


konsoletyper:
Instead of testing `T` classes, I'd just updated `pom.xml` of `tests` modules to accept Java 11 libraries. Since travis-ci configuration uses Java 11, it will be ok.

konsoletyper:
BTW, just did it myself. You can rebase onto master and remove T prefixes from tests

hohwille:
> BTW, just did it myself. You can rebase onto master and remove T prefixes from tests

Awesome. Done. If travis does not give surprises, it should be ready for merge.

PR #483, opened 2020-03-01
support for Thread.UncaughtExceptionHandler

hohwille:
Support for `UncaughtExceptionHandler` in `Thread`. Some libraries are using the Java feature to handle exceptions (e.g. event bus). It would be nice to allow using such lib in TeaVM.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TDefaultUncaughtExceptionHandler.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TDefaultUncaughtExceptionHandler.java
@@ -1,0 +1,29 @@
+/*
+ *  Copyright 2020 by Joerg Hohwiller
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


konsoletyper:
There's no actual implementation for this concept in this PR. You just provide API but don't use in `Thread` implementation.

hohwille:
> There's no actual implementation for this concept in this PR. You just provide API but don't use in Thread implementation.

https://github.com/konsoletyper/teavm/blob/7452146b8b345df27c24f6caa1f59bd11dbc0f06/classlib/src/main/java/org/teavm/classlib/java/lang/TDefaultUncaughtExceptionHandler.java#L26

konsoletyper:
I don't mean that. I mean that this PR does not affect thread behaviour. When thread crashes with uncaught exception, it still does not call this handler

hohwille:
> I don't mean that. I mean that this PR does not affect thread behaviour. When thread crashes with uncaught exception, it still does not call this handler

OK. I see you point. Thanks for the suggestion. Makes perfectly sense. 

I tried what you suggested. Hope that I got it right. Please have a look and review my latest change.

hohwille:
```
[INFO] --- maven-checkstyle-plugin:2.16:check (validate) @ teavm-classlib ---
[INFO] Starting audit...
/home/travis/build/konsoletyper/teavm/classlib/src/main/java/org/teavm/classlib/java/lang/TThread.java:278: Line is longer than 120 characters (found 121).
Audit done.
```

I really love your checkstyle build breaker ;)

PR #488, opened 2020-03-22
Add ability to fine tune transpiler class substitutions.

adamjryan:


konsoletyper:
What is this commit for? What problem is it trying to solve? Please, describe in detail. I know that it improves class renaming mechanism. But why do you actually need it? How are you going to use these improvements?

adamjryan:
The behaviour prior to this change is to enforce quite a strict mapping where if a mapping is defined for a package then all the classes in that package must have a mapping defined as well as all of them applying a prefix if that property is also set. This has a number of shortcomings:
* There may only be a need to implement a single class but all the other classes in the package will also need to be created if they appear in the code.
* If a mapping for a package has been included such as in the teavm classlib artifact then it is not possible to override it should there be problems with missing implementation or bugs, it is all or nothing. There is a slight workaround to hack the classpath ordering to override classes in the classlib jar but obviously it's not ideal.

This change makes it possible to include the teavm classlib mappings and use it selectively so if the default JDK classes transpile without issues then no need to use the reimplementation. This can be done by including just the package/class you need or excluding the ones you don't.
I also think it will let this project focus more on the transpilation and the massive task of implementing a JDK can be separated out as users of the library only need to map what they need and may even be able to get away without it.
Other people can create JDK implementations for bits they want and include it in tandem with the teavm classlib without needing to bother you with pull requests.
I've added the prefix mapping rules for compatibility with the existing behaviour, otherwise I don't think they are that useful.

konsoletyper:
This does not explain to me anything. Can you describe you actual problem? What are you doing? Are you implementing your own Java class library for TeaVM?

adamjryan:
Ok, I want to use the classlib files but they have missing implementation and I want to quickly override them without messing around with classpath's or raising PR's. This change means I can do that by with a mapping rule.

adamjryan:
Does that answer your question?
Also more generally, is the transpiler mapping funtionality open to improvements?

konsoletyper:
I read your answer and understood it. I agree with your reasoning, but I had no chance to look though your commit carefully. As for future changes, you better suggest some changes before working on PR.

adamjryan:
No probs, I'll leave it with you.

PR #498, opened 2020-05-09
497: fix of exporting a wasm function for scala objects

vgordievskiy:
* updated the ClassReader - storing the sourceFile info.
* added ModelUtil functions for checking scala source file.


vgordievskiy:
code example:

object Test {

def main(args: Array[String]): Unit = {
println("Main")
}

@export(name = "hello")
def hello(): Unit = {
println("Hello")
}
}

Error:
Caused by: java.lang.IllegalArgumentException: Function hello already defined in this module

konsoletyper:
It's a shame on Scala compiler developers that they (1) don't bother to add `SYNTHETIC` modifier on a synthesized classes/methods (2) copy annotations to these synthesized methods. I should think a little whether this dirty hack is better or worse than other possible dirty hacks. And the right solution would be to convince Scala developers to *not* to generate class files like these.

core/src/main/java/org/teavm/backend/lowlevel/generate/NameProviderWithSpecialNames.java
```diff
@@ -47,12 +49,14 @@ public String forMethod(MethodReference method) {
 
     private String getSpecialName(MethodReference methodReference) {
         MethodReader method = classSource.resolve(methodReference);
+        ClassReader cls = classSource.get(method.getReference().getClassName());
```
konsoletyper, line 52: I would not spoil compiler core with Scala-specific hacks. There's dedicated class for this called `ScalaHacks`. You can just detect these synthetic Scala classes there and remove `@Export` annotations from their methods.

core/src/main/java/org/teavm/model/util/ModelUtils.java
```diff
@@ -30,9 +30,23 @@
...
 
 public final class ModelUtils {
+
+    private static String scalaMarker = ".scala";
+    private static String scalaInternalMarker = "$";
+
     private ModelUtils() {
     }
 
+    public static Boolean isScalaClass(ClassReader cls) {
```
konsoletyper, line 40: Is there any reason to return `java.lang.Boolean` instead of primitive `boolean`?

core/src/main/java/org/teavm/parsing/Parser.java
```diff
@@ -325,6 +325,10 @@ public ClassHolder parseClass(ClassNode node) {
             }
         }
 
+        if (node.sourceFile != null) {
```
konsoletyper, line 328: Is there any purpose in this `if`?

core/src/main/java/org/teavm/model/util/ModelUtils.java
vgordievskiy: No, there is not any reason) Will change it

core/src/main/java/org/teavm/parsing/Parser.java
vgordievskiy: I will check back it.

core/src/main/java/org/teavm/backend/lowlevel/generate/NameProviderWithSpecialNames.java
vgordievskiy: I've got it. I will redo this solution. Thanks!

core/src/main/java/org/teavm/model/util/ModelUtils.java
vgordievskiy: I've reverted it.

core/src/main/java/org/teavm/parsing/Parser.java
vgordievskiy: I've reverted it.

core/src/main/java/org/teavm/backend/lowlevel/generate/NameProviderWithSpecialNames.java
vgordievskiy: I've update the solution. Now it uses the ScalaHacks.

PR #509, opened 2020-06-11
Adding DOMParser and XMLSerializer

fb71:


ScraM-Team:
Thanks for contributing new bindings!  I can see using both of these browser APIs.

I haven't used the serializeToStream method.  Can you tell us what you've used it for?  MDN suggests it may not be supported in the future: https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer

PR #511, opened 2020-06-13
Hashmap foreach

Ihromant:


konsoletyper:
Won't this PR affect LinkedHashMap order? I think `foreach` method should be overridden for this class as well.

Ihromant:
Added override, removed unused import from HashMap

konsoletyper:
BTW, looks like you copied this code from OpenJDK

konsoletyper:
Also, did you run checkstyle checks? This code looks like improperly formatted

PR #512, opened 2020-06-13
Collectors toMap

Ihromant:


konsoletyper:
Where did you get that code? From OpenJDK? Copying OpenJDK code here violates OpenJDK license.

Ihromant:
Yes, I adapted it from OpenJDK (not copied 1 to 1 due to minor difference). How can I fix this issue? I mean what criteria is for "copied" or "not copied code".

konsoletyper:
Criteria are: you actually got this code from OpenJDK. To fix the issue you should write this code manually from scratch, not looking at the OpenJDK implementation, but only reading spec. Another way to fix is to find any implementation of these methods in a project licensed under Apache license.

PR #513, opened 2020-06-15
Hashmaps foreach

Ihromant:
Thanks for the comments on previous PR.
Now implemented from scratch.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/THashMap.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/THashMap.java
@@ -248,10 +248,10 @@ public class THashMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
                     while (entry != null) {
                         action.accept(entry);
                         entry = entry.next;
+                        if (prevModCount != associatedMap.modCount) {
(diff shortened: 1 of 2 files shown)


classlib/src/main/java/org/teavm/classlib/java/util/THashMap.java
```diff
@@ -237,6 +238,23 @@ private static boolean valuesEq(TMap.Entry<?, ?> entry, TMap.Entry<?, ?> oEntry)
...
+            if (associatedMap.elementCount > 0) {
+                int prevModCount = associatedMap.modCount;
+                for (int i = 0; i < associatedMap.elementData.length; i++) {
+                    HashEntry<K, V> entry = associatedMap.elementData[i];
+                    while (entry != null) {
+                        action.accept(entry);
+                        entry = entry.next;
+                    }
+                }
+                if (prevModCount != associatedMap.modCount) {
```
konsoletyper, line 253: I think this check should be done after every `accept` call. At least, this behaviour will be consistent with default `forEach` implementation from `Iterable` class.

Ihromant: Did it.

PR #518, opened 2020-07-17
Error message if @JSBody with return type does not return

eriklumme:
If a @JSBody-annotated method has a return type (not void), but does not
have a return statement in the script, print an error.

I can't think of a case where someone would actually want the returned value to always be `undefined`, as it would be if the return was missing.

This addition in no way validates the script, and does not catch all cases, but prevents simple errors such as

```
@JSBody(params = "id", script = "document.getElementById(id)")
public static native HTMLElement getElementById(String id);
```

### DIFF
--- a/jso/impl/src/main/java/org/teavm/jso/impl/JSClassProcessor.java
+++ b/jso/impl/src/main/java/org/teavm/jso/impl/JSClassProcessor.java
@@ -550,6 +550,12 @@ class JSClassProcessor {
                 .map(AnnotationValue::getString)
                 .toArray(String[]::new) : new String[0];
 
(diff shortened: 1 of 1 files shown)


konsoletyper:
I won't accept this PR. First, it will give false negative if you write a script something like that: `console.log('return');`. Another example is:

```javascript
if (someCondition()) {
  return 'foo';
}
```

So what you need is to perform control-flow analysis on JS AST. Also, in some cases `undefined` is a value that's expected, so this CFA will produce false positives.

I believe, this should be a warning, not an error. And it should only guard from such trivial cases like you provided. That's why a better behaviour would be: report a warning if return type is not `void` and the entire script body consists of single expression statement.

PR #520, opened 2020-08-01
Stream min max fix

Ihromant:


tests/src/test/java/org/teavm/classlib/java/util/stream/StreamTest.java
```diff
@@ -287,6 +287,27 @@ public void minMax() {
...
+        private final int number;
+        private final int id;
+
+        private BrokenComparable(int number, int id) {
+            this.number = number;
+            this.id = id;
+        }
+
+        @Override
+        public int compareTo(BrokenComparable that) {
```
konsoletyper, line 308: How is this method used?

tests/src/test/java/org/teavm/classlib/java/util/stream/StreamTest.java
```diff
@@ -287,6 +287,27 @@ public void minMax() {
...
         assertFalse(Stream.<Integer>empty().max(Comparator.comparing(x -> x)).isPresent());
+
+        BrokenComparable first = new BrokenComparable(0, 0);
+        BrokenComparable second = new BrokenComparable(0, 1);
+
+        assertEquals(0, Stream.of(first, second).max(Comparator.comparing(x -> x)).get().id);
+        assertEquals(0, Stream.of(first, second).min(Comparator.comparing(x -> x)).get().id);
+    }
+
+    private static class BrokenComparable implements Comparable<BrokenComparable> {
```
konsoletyper, line 298: Does this class really need to be Comparable?

konsoletyper: Is "BrokenComparable" a right name? What about just IntPair or something like this?

konsoletyper (changes requested):
I noticed there's nothing about "values are identical" case in specification, however JDK implementation differs from one in TeaVM.

Ihromant:
There main reason of this fix is that fact that Collextions.max and Stream.max are different in behaviour. Expected (and more convenient) is in case of two possible maximums, return first. Current Stream.max implementation returns second (last).

tests/src/test/java/org/teavm/classlib/java/util/stream/StreamTest.java
Ihromant: The reason of this name is because comparable isn't aligned with possible "equals", where equals will be by id, but compared values will be by value. So in similar cases (I had more complex), we are expecting max to return first found object, but in reality it returned last (JVM in both cases for Collections.max and Stream.max returns first).

Ihromant: So, name means conscious break of "compare-equals" contract for some reason. Can rename it to IntPair.

tests/src/test/java/org/teavm/classlib/java/util/stream/StreamTest.java
Ihromant: It's used in Stream.max(). It can be replaced by similar comparator.

PR #524, opened 2020-09-16
Image data fix

Ihromant:
data property is readonly in ImageData. Here is proper API fix

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/canvas/ImageData.java
+++ b/jso/apis/src/main/java/org/teavm/jso/canvas/ImageData.java
@@ -15,20 +15,30 @@
  */
 package org.teavm.jso.canvas;
 
+import org.teavm.jso.JSBody;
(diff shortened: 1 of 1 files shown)


jso/apis/src/main/java/org/teavm/jso/canvas/ImageData.java
```diff
@@ -15,20 +15,30 @@
...
+import org.teavm.jso.JSBody;
 import org.teavm.jso.JSObject;
 import org.teavm.jso.JSProperty;
 import org.teavm.jso.typedarrays.Uint8ClampedArray;
 
-public interface ImageData extends JSObject {
-    @JSProperty
-    int getWidth();
+public abstract class ImageData implements JSObject {
+    protected ImageData() {
```
konsoletyper, line 24: private?

konsoletyper (changes requested):

PR #532, opened 2020-10-14
Array buffer view create fix

Ihromant:


konsoletyper:
@Ihromant what was the purpose of this PR. What does it "fix"? Looks like everything was okay and this PR just broken things (for example, in our project).

Ihromant:
Hi @konsoletyper ,
I have next code:
DataView oldView = DataView.create(ctx.getImageData(0,  0, WIDTH, HEIGHT).getData()); // ctx is Context2D
Before the fix it was falling, so I used
DataView oldView = DataView.create(ctx.getImageData(0,  0, WIDTH, HEIGHT).getData().getBuffer());
After the fix it started working.
When you review:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/DataView
you can check that there is no constructor from other DataView. I supposed that it was added by mistake, so tried to fix with getting ArrayBuffer and passing it to constructor. Now with reviewing other constructors like:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8ClampedArray/Uint8ClampedArray
I can see that my fix was partially wrong. I suppose that the best fix would be removing creator of DataView from other ArrayBufferView (because there is no such JS method) and fixing the broken constructor that you have.
Sorry for inconvenince.

PR #534, opened 2020-10-25
Template collections fix

Ihromant:
Fixed multiple elements set and map for values with negative hashcode. Simplified code


classlib/src/main/java/org/teavm/classlib/java/util/TTemplateCollections.java
```diff
@@ -201,32 +201,19 @@ public int size() {
             for (T element : data) {
                 Objects.requireNonNull(element);
 
-                int suggestedIndex = element.hashCode() % data.length;
+                int suggestedIndex = Math.abs(element.hashCode()) % data.length;
                 int index = suggestedIndex;
-                boolean found = false;
-                while (index < data.length) {
-                    T existingElement = table[index];
+                while (index < suggestedIndex + data.length) {
```
konsoletyper, line 206: What's the purpose of this change?

Ihromant: Code simplification and unification. Main issue was fixed by Math.abs(hashCode), but considered that it would be better to have fewer and clearer lines of code.

konsoletyper: Please, revert this. I intentionally unswitched this loop manually to make it faster.

Ihromant: Ok, will do it.

Issue #519, opened 2020-07-21
subMap does not return empty maps correctly


And-rea-Q:
Small test case:

```java 
    public static void main(String[] args) {

        TreeMap<Integer, Integer> myTreeMap = new TreeMap<Integer, Integer>();
        myTreeMap.put(1, 1);
        myTreeMap.put(3, 15);
        myTreeMap.put(4, 20);
        myTreeMap.put(6, 13);
        myTreeMap.put(10, 119);

        System.out.println(printMap(myTreeMap.subMap(0, 0)));
        System.out.println(printMap(myTreeMap.subMap(7, 9)));
        System.out.println(printMap(myTreeMap.subMap(3, 9)));
        System.out.println(printMap(myTreeMap.subMap(10, 29)));
        System.out.println(printMap(myTreeMap.subMap(29, 100)));
    }
    static <K, V>
    String printMap(Map<K, V> myMap) {
        StringBuilder sb = new StringBuilder("");
        for (Map.Entry<K, V> entry : myMap.entrySet()) {
            sb.append(entry.getKey());
            sb.append(" = ");
            sb.append(entry.getValue());
            sb.append(", ");
        }
        return sb.toString();
    }
```
expected output:
```
1: 
2: 
3: 3 = 15, 4 = 20, 6 = 13, 
4: 10 = 119, 
5: 
```

gotten output:
```
1: 1 = 1, 3 = 15, 4 = 20, 6 = 13, 10 = 119,
2: 10 = 119, 
3: 3 = 15, 4 = 20, 6 = 13, 
4: 10 = 119, 
5: 
```

ScraM-Team:
Thanks for filing this report, I will take a look.

ScraM-Team:
I get the same result.  In the browser console:

```
1 = 1, 3 = 15, 4 = 20, 6 = 13, 10 = 119, 
10 = 119, 
3 = 15, 4 = 20, 6 = 13, 
10 = 119, 

```

Ihromant:
I suppose that we can use implementation from Apache Harmony to fix this issue.
https://github.com/pulseenergy/harmony-treemap
here is example of such implementation.
@konsoletyper what do you think?

PR #535, opened 2020-10-25
Treemap submap fix

Ihromant:
Adapted TreeMap implementation from https://github.com/pulseenergy/harmony-treemap here.
There are also lots of tests that can be moved to our tests.
This implementation fixes https://github.com/konsoletyper/teavm/issues/519 issue.


konsoletyper:
Won't approve this. First, I insisted to not use Apache Harmony map, it was possible just to fix one particular issue in existing implementation. Second, tests are not written properly. Please, see how existing library test are written (see `tests` directory). You should also run tests with special flags passed to JUnit, see `TeaVMTestRunner` implementation for details. For example, my local configuration is following: `-ea -Dteavm.junit.target=target/js-tests -Dteavm.junit.js.runner=htmlunit -Dteavm.junit.threads=6 -Dteavm.junit.minified=false -Dteavm.junit.optimized=false -Dteavm.junit.js.decodeStack=false`

Ihromant:
Thanks. Will try.
I suppose that this configuration should be moved to readme.md with how-to run tests.

PR #554, opened 2021-01-11
Double tostring without long

Ihromant:
Hi @konsoletyper 
Here is a prototype of PR for Double::toString improvement.

**Motivation:** I had a large svg generation in my project, so used lots of doubles to generate polygons. And even on approx 200 polygons JS showed bottleneck in performance checking. I replaced it to int (works for me), but investigated the reason. The reason according to profiler, was in double to string serialization, especially with slow long operations. In my opinion, it's wrong to rely by double (native JS type) to long (emulated JS type). So, today I got the idea about how it can be implemented without longs (based on Double::toString/Number::toString documentation in Java and JS).
This is not a final PR definitely, lots of bugfixes, optimizations (like one resizing in the beginning and operating directly with buffer array, avoiding string operations) should be done. But to continue, I need few answers from you:
1. Is it worth to finish this PR? Or it will be rejected?
2. If there are conditions when this PR can be accepted, what are they?
3. In case when first 2 answers are positive, how to write benchmarks which would compare old and new implementation?


konsoletyper:
Motivation to not accept this PR:

1. JS toString may be different in small details compared to Java toString
2. JS toString implementation is definitely different compared to Java implementation in TeaVM, which will produce inconsistency between backends

I suggest you to take alternative approach.

1. Bring heavy computation to separate Web worker.
2. Use float or int with fixed point
3. Try to define something like DoubleUtil class in your project which uses the same platform defection mechanism that you use in this PR
4. If you can share you project, please, provide instructions how to use it and get performance bottleneck, so I could investigate it myself and think if I can improve Double.toString performance.

Ihromant:
https://github.com/Ihromant/hexagon
Here is simple project where I extracted bottleneck functionality, cleaned it up from streams etc.
On my newer laptop main method runs approx 330 ms, on my older - approx 450 ms in Chrome in dev mode.
Profiler shows that most of this time is consumed by Double::toString and long class initializer.

Ihromant:
@konsoletyper any updates on Double.toString?
https://tc39.es/ecma262/#sec-numeric-types-number-tostring - here is implementation requirements for toString in ECMAScript
https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#toString-double- - here is implementation requirements for toString in Java
From this two desciptions follows that this two implementation are convertible and approach should work without long emulation usage (which is very slow).

konsoletyper:
I'm not going to fix that. I proposed you some alternative solutions.

Ihromant:
I already understand that Double.toString is slow, so I'm avoiding it where possible, but other users can suffer from implicit usage of Double.toString which can happen everywhere (string concat, autogenerated toString etc.).
Still, I'm suggesting at some time fix this performance bottleneck, even with minor difference in some corner cases. My approach is one which can be used for this purpose. GWT (and probably J2CL) just does value + "" to emulate Double.toString and nobody cares.

Ihromant:
Created issue https://github.com/konsoletyper/teavm/issues/560 to reflect this behavior.

PR #555, opened 2021-01-26
#451: add missing methods to java.util.Math (required by java.time)

hohwille:
add missing methods to Math required by `java.time` (for #451).
As discussed in https://github.com/m-m-m/tvm/issues/2 I am creating a temporary external extension module for `java.time` support for TeaVM so I am not blocked with TeaVM.
My plan is to do small PRs with changes in TeaVM that make sense, are consistent and can therefore (hopefully) be merged and included in upcoming releases.
(I can not easily change existing `T*` classes from classlib outside TeaVM in my lib as the classloader order (that is kind of random) then decides whether the original class from TeaVM or my alternative implementation is used. Therefore existing classes like `TMath` should be extended directly in TeaVM project).

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
@@ -21,8 +21,6 @@ import org.teavm.interop.Import;
 import org.teavm.interop.NoSideEffects;
 import org.teavm.interop.Unmanaged;
 
-import jdk.internal.HotSpotIntrinsicCandidate;
-
 @NoSideEffects
 public final class TMath extends TObject {
     public static final double E = 2.71828182845904523536;
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
@@ -21,6 +21,8 @@ import org.teavm.interop.Import;
 import org.teavm.interop.NoSideEffects;
 import org.teavm.interop.Unmanaged;
 
+import jdk.internal.HotSpotIntrinsicCandidate;
+
 @NoSideEffects
 public final class TMath extends TObject {
     public static final double E = 2.71828182845904523536;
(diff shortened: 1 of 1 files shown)


hohwille:
Any feedback on this PR yet? Merging this would really help me to get out of the blockade I am currently facing with TeaVM.
My dream would be that after merging this PR a new release of TeaVM gets released (ideally to maven central, but bintray would also help). Since more than one year I am unable to push out a working demo of https://github.com/m-m-m/ui-demo/ build with TeaVM. It works when I build TeaVM locally from this PR. Surely you can say "what the fuss" and "this is your problem and not ours" but it could also be a nice promotion for TeaVM to see great dreams and demos comming true with the help of TeaVM.

konsoletyper:
For what purpose you need these methods?

hohwille:
As described above for `java.time` support in the browser.

konsoletyper:
I believe this PR is not necessary for your java.time implementation. AFAIR, you use threeten implementation which does not rely on java 8 method, instead it comes with `Jdk8Methods` class which contains implementation for these methods.

hohwille:
Surely I can bypass `Math` in the T* classes emulating `java.time`.
So if it is a problem for you to merge this I have to add more workarounds to make TeaVM compile my code...

Ihromant:
Actually, methods @hohwille provided are in JDK 8-9, so you can merge this PR as usual addition of missing methods.

classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
```diff
@@ -153,6 +153,114 @@ public static long floorMod(long a, long b) {
...
+        if (value > Integer.MAX_VALUE || value < Integer.MIN_VALUE) {
+            throw new ArithmeticException("Calculation overflows an int: " + value);
+        }
+        return (int) value;
+    }
+
+    public static int addExact(int a, int b) {
+        int sum = a + b;
+        if ((a ^ sum) < 0 && (a ^ b) >= 0) {
+            throw new ArithmeticException("Addition overflows an int: " + a + " + " + b);
```
konsoletyper, line 208: I consider not reporting a and b here, to improve performance and code size

classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
```diff
@@ -153,6 +153,114 @@ public static long floorMod(long a, long b) {
...
+        }
+        return sum;
+    }
+
+    public static long addExact(long a, long b) {
+        long sum = a + b;
+        if ((a ^ sum) < 0 && (a ^ b) >= 0) {
+            throw new ArithmeticException("Addition overflows a long: " + a + " + " + b);
```
konsoletyper, line 216: I consider not reporting a and b here, to improve performance and code size

And so on in similar cases

classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
```diff
@@ -153,6 +153,114 @@ public static long floorMod(long a, long b) {
...
+    public static int toIntExact(long value) {
+        if (value > Integer.MAX_VALUE || value < Integer.MIN_VALUE) {
+            throw new ArithmeticException("Calculation overflows an int: " + value);
+        }
+        return (int) value;
+    }
+
+    public static int addExact(int a, int b) {
+        int sum = a + b;
+        if ((a ^ sum) < 0 && (a ^ b) >= 0) {
```
konsoletyper, line 207: This means

> `a` and `b` have same sign (both either positive or negative), whereas sum has different sign. We use ^ only do check higher bit (which would be a sign), other bits are not interesting here.

It's a good place to add a comment

konsoletyper:
This PR lacks tests

PR #569, opened 2021-04-23
Simple split

Ihromant:
Simple split implementation when we're passing one-character regexp implemented according to https://youtu.be/YgGAUGC9ksk?t=2426


konsoletyper (commented):
One thing that you should realize that even with this method String.split will have huge cost due to need to compile whole regexp engine (at compile time it's unclear if first parameter passed to `split` method is one-character constant string). Personally I avoid using regular expressions and `String.split` method. Unfortunately, Java was not designed with AOT in mind and `String.split` is one example of this. So it's better to use StringUtils from apache-commons-lang or StringTokenizer class.

Ihromant:
@konsoletyper my idea was about eliminating pattern presence in cases when String argument is constant, for instance: "a,b,c".split(","). If you are claiming that whole pattern will be compiled anyway, then it adds just minor performance improvement. I thought that TeaVM compiler would be able to detect that the branch with Pattern is not reachable and don't produce additional JS. In my case it's +100Kb of JS.

Issue #579, opened 2022-01-30
TouchEvent missing in JSO DOM events?


coder111111:
Hi,

First let me tell you I'm seriously impressed with TeaVM. I've been experimenting with TeaVM and Three.js for last couple of months and the results are pretty amazing.

That being said, I wanted to see how my website works on a mobile device, and I'm unable to work with TouchEvents, mainly because the class itself seems to be not implemented in TeaVM. More details about touch events:

https://www.w3schools.com/jsref/obj_touchevent.asp

Would it be possible to add it? Or would you merge a PR if I were to add it myself?

Sincerely,
--Coder

konsoletyper:
You can write corresponding event type right in your code base, it's not really necessary to have them in TeaVM code base. It's just a matter of convenience: new developers won't need to re-write these interfaces on their side and re-use built-in one.

ScraM-Team:
Coder,

> First let me tell you I'm seriously impressed with TeaVM. I've been experimenting with TeaVM and Three.js for last couple of months and the results are pretty amazing.

Can you share what you've been working on?  We're always interested in hearing what people are doing with TeaVM.

> Would it be possible to add it? Or would you merge a PR if I were to add it myself?

Thanks for the PR!  Please keep at it, this is useful for many TeaVM users.  The more mappings available in TeaVM, the easier it is for new users to get started.

PR #580, opened 2022-01-30
Add Touch Event. Fixes issue #579

coder111111:
Hi,

I wanted to play with touch events today, so I added them myself. Please review code and merge upstream if you like it.

--Coder

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/dom/events/Touch.java
+++ b/jso/apis/src/main/java/org/teavm/jso/dom/events/Touch.java
@@ -1,0 +1,62 @@
+/*
+ *  Copyright 2014 Alexey Andreev.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.jso.dom.events;
+
+import org.teavm.jso.JSObject;
+import org.teavm.jso.JSProperty;
+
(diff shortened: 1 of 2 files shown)


jso/apis/src/main/java/org/teavm/jso/dom/events/Touch.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.teavm.jso.JSObject;
+import org.teavm.jso.JSProperty;
+
+// From https://developer.mozilla.org/en-US/docs/Web/API/Touch
+
+public interface Touch extends JSObject {
+
+    @JSProperty
+    long getIdentifier();
```
konsoletyper, line 26: This won't work. JavaScript does not provide int64 type (or at least, did not provide by the time touch event spec was written). I'd replaced this with `int`, but you should test on real use cases if `int` fits well or you need `double`.

jso/apis/src/main/java/org/teavm/jso/dom/events/TouchEvent.java
```diff
@@ -0,0 +1,46 @@
...
+ *  limitations under the License.
+ */
+package org.teavm.jso.dom.events;
+
+import org.teavm.jso.JSProperty;
+import org.teavm.jso.core.JSArray;
+
+// From https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent
+
+public interface TouchEvent extends Event {
```
konsoletyper, line 23: Ok, you created an event class. But how one's to use it. What about declaring corresponding event target interface, like I did it with MouseEventTarget, for example?

jso/apis/src/main/java/org/teavm/jso/dom/events/TouchEvent.java
```diff
@@ -0,0 +1,46 @@
...
+    boolean getAltKey();
+
+    @JSProperty
+    boolean getMetaKey();
+    
+    @JSProperty
+    JSArray<Touch> getChangedTouches();
+    
+    @JSProperty
+    JSArray<Touch> getTargetTouches();
```
konsoletyper, line 41: 1. Is it really necessary to expose this property as `JSArray`? What about just `Touch[]`?
2. If you still want JS native type, consider `JSArrayReader<Touch>` here

konsoletyper (changes requested):

Ihromant:
This PR can be closed.

PR #582, opened 2022-02-12
Fixed invalid JSBody annotation in IDBObjectStore, fixed IDBFactory compatibility check

LAX1DUDE:
The annotation was missing a 'params' definition in the constructor which prevented the IndexedDB JSO API from successfully compiling to JavaScript in the current version of TeaVM.

**I do not know if this was the ideal way to patch this function, feel free to reject this request and patch it differently if the 30 second change I have provided here was not the correct way to fix this problem**

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/indexeddb/IDBObjectStore.java
+++ b/jso/apis/src/main/java/org/teavm/jso/indexeddb/IDBObjectStore.java
@@ -36,7 +36,7 @@ public abstract class IDBObjectStore implements JSObject, IDBCursorSource {
         }
     }
 
-    @JSBody(script = "return this;")
+    @JSBody(params = "obj", script = "return obj;")
     private native String[] unwrapStringArray(JSObject obj);
 
     @JSProperty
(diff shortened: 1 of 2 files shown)


LAX1DUDE:
I discovered `IDBFactory` had a broken compatibility check, it was throwing an 'unsupported' exception when `window.indexedDB` is **not undefined** instead of when it is undefined, I added it to this same pull request as the `IDBObjectStore` patch.

I'm the guy porting Minecraft to TeaVM if you forgot. I am trying to add singleplayer via web workers and using IndexedDB to persistently store the gigabytes of worlds so I'm pretty much gonna end up finding myself testing the entire TeaVM IndexedDB API so I'll make sure to add more commits to this request or open more if I discover any more inconsistencies

jso/apis/src/main/java/org/teavm/jso/indexeddb/IDBObjectStore.java
```diff
@@ -36,7 +36,7 @@
         }
     }
 
-    @JSBody(script = "return this;")
+    @JSBody(params = "obj", script = "return obj;")
     private native String[] unwrapStringArray(JSObject obj);
```
konsoletyper, line 40: I think this method should also be marked as 'static'

konsoletyper (changes requested):

jso/apis/src/main/java/org/teavm/jso/indexeddb/IDBObjectStore.java
LAX1DUDE: Oh I didn't catch that, thanks for pointing it out

PR #586, opened 2022-05-06
add `Objects.requireNonNullElse` and `Objects.requireNonNullElseGet`

ulugbekna:


konsoletyper:
Is it copy-paste from OpenJDK? Then I can't accept this PR due to licensing issues.

ulugbekna:
I consulted with openjdk but there aren’t that many ways to implement this, are there?

PR #589, opened 2022-06-04
Use ASM9

Thihup:


konsoletyper:
What is the purpose of this commit?

Thihup:
First step to allow building with JDK 17

konsoletyper:
But why did you replace all these `Opcodes.ASM7`? Did you have any troubles with ASM API compatibility?

PR #599, opened 2022-07-11
minor refactorings

ulugbekna:
Was reading the code and noticed some minor cleanups possible

- TeaVMTool: re-use already created file
- TeaVMTool: use a single try-with-resources
- TeaVMTool: `resolveTransformers()` doesn't need to take as an argument variable that it can access itself
- TeaVMTool: handle target directory creation error


kirillp:
This change breaks build
Before this "mvn package" invocation did a new build. Now it prints "Target directory could not be created" and does nothing

The mkdirs() returns false when the directory already exists and it is not a blocker to compile.
Undoing the commit fixes the build

PR #601, opened 2022-07-14
update iml's to java 8

ulugbekna:
otherwise, one has to reconfigure module settings to use java 8, and that gets out of sync on reimport

I'd like to find some time to understand why we need these `.iml`s in the first place (and maybe remove them)


konsoletyper:
The reason is: IDEA is unable to import Maven modules as IDEA plugin modules. This requires to leave these `.iml` files in VCS.

ulugbekna:
1. We could still upgrade to java 8 right? 
2. What do you think of this https://youtrack.jetbrains.com/issue/IDEA-65130 -- seems fixed?

PR #612, opened 2022-08-13
Moved next origin/bound implementations from TThreadLocalRandom to TRandom.

iamdudeman:
This moves the implementations of the origin/bounds next methods from TThreadLocalRandom to TRandom, replacing the implementations I previously added. These implementations were more accurate and had appropriate input validations in place. 

Thank you @Ihromant for pointing out the better implementations in TThreadLocalRandom. I am still very new to this codebase :)


classlib/src/main/java/org/teavm/classlib/java/util/TRandom.java
```diff
@@ -75,15 +75,54 @@ public int nextInt(int n) {
...
+                }
+            }
+        }
     }
 
     public long nextLong() {
         return ((long) nextInt() << 32) | nextInt();
     }
 
+    public long nextLong(long bound) {
```
iamdudeman, line 98: Would it make sense to use a similar implementation to `int nextInt(int n)` here instead of what was in the private method?

```
public long nextLong(long bound) {
        if (bound <= 0) {
            throw new IllegalArgumentException();
        }
        return (long) (nextDouble() * bound);
    }
```

Ihromant: No, it's not enough of bytes (52 if I recall correctly) in mantissa of double to cover all possible longs. You're cutting most of possible results by this implementation.

Ihromant (approved):

PR #637, opened 2022-11-08
Add ExecutionException to standard library.

colorizenl:


classlib/src/main/java/org/teavm/classlib/java/util/concurrent/TExecutionException.java
```diff
@@ -0,0 +1,27 @@
+/*
+ *  Copyright 2014 Alexey Andreev.
```
konsoletyper, line 2: Obviosly, it's 2022 and your name

PR #639, opened 2022-11-17
Crypto/UUID

Ihromant:
Use crypto module for generating UUID in JS code


Ihromant:
@konsoletyper added check for crypto support and introduced crypto object

jso/apis/src/main/java/org/teavm/jso/crypto/Crypto.java
```diff
@@ -0,0 +1,47 @@
...
+public abstract class Crypto implements JSObject {
+    @JSBody(script = "return window.crypto != null;")
+    public static native boolean isSupported();
+
+    @JSBody(script = "return window.crypto;")
+    public static native Crypto current();
+
+    public abstract String randomUUID();
+
+    public abstract void getRandomValues(Int8Array arr);
```
konsoletyper, line 36: It's also good to add methods that take Java array with `@JSByRef` annotation.

PR #641, opened 2022-12-03
Class forname/message notfound

Ihromant:


Ihromant:
@konsoletyper

konsoletyper:
The reason I don't merge this PR is efficiency. This additional message makes resulting JavaScript bigger. According to my experience, exception messages rarely make development easier, I always need whole stack with variables on stacks, etc. If developer wants to know what particular class caused the problem, they can put breakpoint and see it from locals.

Ihromant:
Many thanks for explanation. I wanted to know the reason.

PR #646, opened 2022-12-08
Jso/path2d

Ihromant:


konsoletyper:
Should corresponding methods be added to RenderingContext2D?

PR #650, opened 2023-01-20
Base64 implementation

Ihromant:
Added Base64.Encoder, Base64.Decoder together with fix for bug found in Base64Impl.


konsoletyper:
I merged the PR, but tests turned out to be broken. Please, run tests before creating PR

Issue #655, opened 2023-02-05
java.util.Date: Year +1900 to Match Java API


noblemaster:
The `year` for the `java.util.Date` is considered `year -  1900` in Java. The corresponding `TDate` class is mostly correct, but has `+ 1900` missing at the following line:

https://github.com/konsoletyper/teavm/blob/bdbce01b9a3c45cc7233411519d7d6e7076e6b81/classlib/src/main/java/org/teavm/classlib/java/util/TDate.java#LL95C30-L95C30

 - Wrong: `return (long) JSDate.UTC(year, month, date, hrs, min, sec);`
 - Correct: `return (long) JSDate.UTC(year + 1900, month, date, hrs, min, sec);`

EDIT: I had another look. I was looking at the wrong file. This should just be a small fix as outlined above.

PR #656, opened 2023-02-08
Bugfix: "+1900" should be added when creating a date via JSDate.UTC.

noblemaster:
Fixes https://github.com/konsoletyper/teavm/issues/655


konsoletyper:
Can you please provide a test? See `tests` folder.

noblemaster:
Indeed, I should have done that from the start. I updated `DateTest.java` which will fail (wrong year) unless the fix is applied.

PR #658, opened 2023-02-12
Bugfix: fixes variable name generator to prevent 'do' from being generated.

noblemaster:
Prevents variable names `do` from being generated which is a reserved keyword in `JavaScript`. Any 2-letter variable names will now work. Fixes https://github.com/xpenatan/gdx-teavm/issues/63 

Note, this will still cause problems for 3+ letter variable names (although 3-letter are probably unlikely to be generated). Maybe it makes sense to remove all `aeiou` and potentially `OU` (e.g. `URL`) as well?


konsoletyper:
I don't think it's a good idea to avoid the entire letter to avoid just couple of keywords. I believe there was a code that blacklists certain words in case of global identifiers (i.e. method and class names). Also, object names (not keywords) aren't a problem, since TeaVM does not address them by identifier, but via special 'global' object.

noblemaster:
Alright, I re-wrote it with the keywords.

noblemaster:
Moved to a separate branch: https://github.com/konsoletyper/teavm/pull/665

PR #659, opened 2023-02-12
Reflect record components

Ihromant:
Initial implementation of Metaprogramming support for RecordComponents.
I did this implementation by analyzing the flow for everything in teavm, so it might be something important missing.


konsoletyper:
Sorry, I can't merge this PR, since it affects IR, which is a crucial part of TeaVM and each change in it should be carefully considered. Also, I doubt if it's a good idea to introduce something in IR just for the needs of single module. Perhaps, IR should be refactored to be able to do this.

Ihromant:
Got it. I considered it as eventual simplification for record serialization-deserialization, but I suppose that I'll be able to overcome it somehow by analyzing method structure of record classes (I have flag, so it will help I think). Still, if refactoring will happen after some time, it would be great to support newest Java versions.

Ihromant:
One more question: there is no constructors in ReflectClass. Is it possible to add it there or it will contradict current IR?

konsoletyper:
There's no contradiction, but aren't constructors available as `<init>` methods?

PR #660, opened 2023-02-13
Xhr addition fixes

Ihromant:
Added methods for XMLHttpRequest, set it to be event target and minor fixes according to API


classlib/src/main/java/org/teavm/classlib/java/net/impl/TXHRURLConnection.java
```diff
@@ -86,7 +86,7 @@ private void performRequestIfNecessary() {
     @Async
     private native Boolean performRequest();
     private void performRequest(AsyncCallback<Boolean> callback) {
-        xhr.setOnReadyStateChange(() -> {
+        xhr.setOnReadyStateChange(e -> {
```
konsoletyper, line 89: So, this commit breaks backward compatibility. Can you please try to fix that? For example, leave old method and introduce the new one instead of changing signature?

konsoletyper (changes requested):

classlib/src/main/java/org/teavm/classlib/java/net/impl/TXHRURLConnection.java
Ihromant: It's not backward compatibility in my opinion, but initially wrong implementation according to https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readystatechange_event (it should be not runnable, but  generic event consumer). Still to remove native call in my code, I just need to inherit XHR from EventTarget. It's up to you to decide whether to leave incorrect implementation for backward compatilibity or change it. Adding requested changes in next commit.

konsoletyper:
Since JS is dynamically typed, it's ok to pass callback that takes other number of parameter, than expected by the API. It's often used by those, who design API, to make changed backward compatible, e.g. adding parameters to callbacks, which is binary compatible in this case. Older XHR specs declared callback as a zero-parameter function. In Java requirements are stronger, so to keep binary compatible you should use another methods, e.g. method overloading.

jso/apis/src/main/java/org/teavm/jso/ajax/XMLHttpRequest.java
```diff
@@ -54,6 +57,27 @@ public abstract class XMLHttpRequest implements JSObject {
     @JSProperty("onreadystatechange")
     public abstract void setOnReadyStateChange(ReadyStateChangeHandler handler);
```
konsoletyper, line 58: I would overload this method to take `EventListener` as well.

Ihromant: Yes, that would work, thanks for the advice.

Ihromant:
By the way what specs are you using as well for JS code?

konsoletyper:
Usually MDN

PR #662, opened 2023-02-15
Bugfix graph splitter

noblemaster:
This fix makes sure only the actual number of nodes are copied (probably same as before) and updates the pointer. 
- Fixes: https://github.com/konsoletyper/teavm/issues/657
 - Probably Fixes (Issue appears related): https://github.com/xpenatan/gdx-teavm/issues/65

EDIT: this solves the problem for my application. It now compiles and runs without problems.


noblemaster:
I also added a test to show the solution works based on the issue reported.

konsoletyper:
I made a better fix

PR #665, opened 2023-02-15
Bugfix: fixes variable name generator to prevent 'do' from being generated.

noblemaster:
Prevents variable names `do` from being generated which is a reserved keyword in JavaScript. The test case will fail unless the bugfix is applied.

Continuance of PR (separate branch): https://github.com/konsoletyper/teavm/pull/658


konsoletyper:
Made my own fix

PR #669, opened 2023-02-26
Null object

Ihromant:


konsoletyper:
What is the purpose of these methods? Why not using `null` and `== null` from Java?

Ihromant:
Ok, I agree, null works everywhere where needed.

PR #674, opened 2023-03-08
Bitset stream

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/util/TBitSet.java
```diff
@@ -511,6 +512,14 @@ public String toString() {
         return sb.toString();
     }
 
+    public TIntStream stream() {
```
konsoletyper, line 515: This implementation is not very efficient, since it tests each bit. Instead it could:

1. Skip elements which are entirely 0
2. Skip all unset bits, there are algorithms that efficiently handle this.

See `nextSetBit`.  I propose instead of this return subclass of `TSimpleIntStreamImpl` with `next` method overridden, see `TRangeIntStream` for example. Then this implementation could just call `nextSetBit` each time, as people usually do to iterate over BitSet.

konsoletyper (changes requested):

PR #678, opened 2023-03-24
Fix TAbstractStringBuilder for MIN_VALUE.

tommyettinger:
`Long.MIN_VALUE` and `Integer.MIN_VALUE` currently are appended as `-0`, instead of `-9223372036854775808` and `-2147483648`, respectively. This fixes the behavior for those values, and keeps other values how they are currently appended. I haven't tried radices other than 10, because the tests only cover radix 10; I added tests for the minimum values to augment the existing tests for the maximum values.

The reason the behavior was wrong before had to do with the oddball behavior of `-Long.MIN_VALUE`, which is the same as `Long.MIN_VALUE` (both negative, and this is also true for Integer). This always works with values significant enough to go into the second branch as negative numbers, which allows `-9223372036854775808` to be used as-is, and all positive numbers fit between `0` and `-9223372036854775808`.

I ran `gradlew check` and it said there were no failing tests or checkstyle violations, but it had some internal Exception that I couldn't find any more info on. I don't know if this is something I can or should change.

```
> Task :tests:test
java.lang.RuntimeException
	at org.teavm.vm.VMTest.throwException(VMTest.java:480)
	at org.teavm.vm.VMTest.lambda$catchExceptionFromLambda$0(VMTest.java:74)
	at org.teavm.vm.VMTest.catchExceptionFromLambda(VMTest.java:75)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.teavm.junit.TeaVMTestRunner$SimpleMethodRunner.run(TeaVMTestRunner.java:784)
	at org.teavm.junit.TeaVMTestRunner.runInJvm(TeaVMTestRunner.java:672)
	at org.teavm.junit.TeaVMTestRunner.runChild(TeaVMTestRunner.java:453)
	at org.teavm.junit.TeaVMTestRunner.run(TeaVMTestRunner.java:333)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:108)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:40)
	at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:60)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:52)
	at jdk.internal.reflect.GeneratedMethodAccessor5.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at jdk.proxy1/jdk.proxy1.$Proxy2.processTestClass(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker$2.run(TestWorker.java:176)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.executeAndMaintainThreadName(TestWorker.java:129)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:100)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:60)
	at org.gradle.process.internal.worker.child.ActionExecutionWorker.execute(ActionExecutionWorker.java:56)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:113)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:65)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
> Task :tests:check
BUILD SUCCESSFUL in 12m 56s
```

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TAbstractStringBuilder.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TAbstractStringBuilder.java
@@ -148,16 +148,16 @@ class TAbstractStringBuilder extends TObject implements TSerializable, TCharSequ
 
     protected TAbstractStringBuilder insert(int target, long value, int radix) {
         boolean positive = true;
-        long isMin;
+        boolean isMin;
         if (value < 0) {
             positive = false;
             value = -value;
-            isMin = (value & -value) >> 63 | 1L;
+            isMin = value == TLong.MIN_VALUE;
         }
         else {
-            isMin = 0L;
+            isMin = false;
         }
-        if (value < radix && isMin == 0L) {
+        if (value < radix && !isMin) {
             if (!positive) {
                 insertSpace(target, target + 2);
                 buffer[target++] = '-';
@@ -168,7 +168,7 @@ class TAbstractStringBuilder extends TObject implements TSerializable, TCharSequ
         } else {
             int sz = 1;
             long pos = -1L;
-            value *= isMin;
+            value = -value;
             while (pos * radix < pos && pos * radix >= value) {
                 pos *= radix;
                 ++sz;
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TAbstractStringBuilder.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TAbstractStringBuilder.java
@@ -99,11 +99,16 @@ class TAbstractStringBuilder extends TObject implements TSerializable, TCharSequ
 
     TAbstractStringBuilder insert(int target, int value, int radix) {
         boolean positive = true;
+        boolean isMin;
         if (value < 0) {
             positive = false;
             value = -value;
+            isMin = value == TInteger.MIN_VALUE;
         }
-        if (value < radix) {
+        else {
+            isMin = false;
+        }
+        if (value < radix && !isMin) {
             if (!positive) {
                 insertSpace(target, target + 2);
                 buffer[target++] = '-';
@@ -112,13 +117,15 @@ class TAbstractStringBuilder extends TObject implements TSerializable, TCharSequ
             }
             buffer[target++] = TCharacter.forDigit(value, radix);
         } else {
-            int pos = 1;
+            int pos = -1;
+            value = -value;
             int sz = 1;
-            int posLimit = TInteger.MAX_VALUE / radix;
-            while (pos * radix <= value) {
+            int posLimit = TInteger.MIN_VALUE / radix;
+
+            while (pos * radix >= value) {
                 pos *= radix;
                 ++sz;
-                if (pos > posLimit) {
+                if (pos < posLimit) {
                     break;
                 }
             }
@@ -129,7 +136,7 @@ class TAbstractStringBuilder extends TObject implements TSerializable, TCharSequ
             if (!positive) {
                 buffer[target++] = '-';
             }
-            while (pos > 0) {
+            while (pos < 0) {
                 buffer[target++] = TCharacter.forDigit(value / pos, radix);
                 value %= pos;
                 pos /= radix;
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TAbstractStringBuilder.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TAbstractStringBuilder.java
@@ -99,6 +99,10 @@ class TAbstractStringBuilder extends TObject implements TSerializable, TCharSequ
 
     TAbstractStringBuilder insert(int target, int value, int radix) {
         boolean positive = true;
+        // TInteger.MIN_VALUE has special behavior.
+        // Because MIN_VALUE == -MIN_VALUE, and MIN_VALUE < 0, assigning:
+        //   value = -value;
+        // Will not make value negative when it is MIN_VALUE.
         boolean isMin;
         if (value < 0) {
             positive = false;
@@ -107,6 +111,9 @@ class TAbstractStringBuilder extends TObject implements TSerializable, TCharSequ
         } else {
             isMin = false;
         }
+        // Here, value is always either non-negative, or MIN_VALUE .
+        // By checking isMin specifically, we can avoid the single-digit
+        // path for MIN_VALUE .
         if (value < radix && !isMin) {
             if (!positive) {
                 insertSpace(target, target + 2);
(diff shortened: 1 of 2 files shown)


konsoletyper:
I think that this implementation with negatives is a bit non-obvious. Can you use divideUnsigned/remainderUnsigned/compareUnsigned instead? I just added implementation of compareUnsigned.

tommyettinger:
I've never used the `xyzUnsigned()` methods in Java 8+ because I've usually had to stay compatible with lower Java versions, which didn't have any of those. I could add comments, but from what I understand, the source for TAbstractStringBuilder isn't visible to users, so the comments would be for other TeaVM devs. I'm pretty sure the "non-obvious" implementation is how this type of operation is supposed to be done to avoid issues with MIN_VALUE, and I don't find operations with `compareUnsigned()` "obvious" at all...

I'll comment this heavily, because I don't feel confident I can rewrite this to use `xyzUnsigned()` and have it remain correct.

konsoletyper:
> I've never used the xyzUnsigned() methods in Java 8+ because I've usually had to stay compatible with lower Java versions, which didn't have any of those. 

I see, but compatibility with Java versions lower than 11 is not the case, so you CAN and SHOULD use necessary tools.

> I'll comment this heavily, because I don't feel confident I can rewrite this to use xyzUnsigned() and have it remain correct.

If it's not so important for you be contribute a PR, I can close it and fix the issue myself. If you want still to merge this PR, please, use `xyzUnsigned` methods.

tommyettinger:
I'm not sure why you insist on requiring xyzUnsigned methods; what do they do that the approach here doesn't? Whatever, it doesn't matter.

I was only PRing this because I wanted to test the waters before making a larger commitment to the formatting code (and because this is a dumb bug). It looks like I won't be contributing at all to TeaVM now. I won't be recommending TeaVM for cross-platform apps until the library is more compatible. I hope it gets there, eventually, but I won't be contributing to that process.

Consider being more flexible in the future.

konsoletyper:
> I'm not sure why you insist on requiring xyzUnsigned methods; what do they do that the approach here doesn't? Whatever, it doesn't matter.

Sorry, it's hard to explain. I think unsigned operations feel more natural here. If I wrote the code in the language which supports unsigned int/long types, I would use them for representing the number being converted, simply because the concept of "sign" makes no sense there, as we separate the given number into the sign and "remaining part" in the very beginning of the method. Working around the issue by using negative number looks like a dirty hack. Since there's a way to support concept of "unsigned integer" types using these xyzUnsigned methods, it looks more natural to use them, in order to reduce mental load on a person who reads the code. Another benefit from using unsigned methods is the possibility to avoid `isMin` flag.

>  won't be recommending TeaVM for cross-platform apps until the library is more compatible.

Can you please explain what you mean?

> Consider being more flexible in the future.

Sorry, but I'm concerned about code quality here. For me it was not hard to fix the issue myself, I just want to help people to become contributors. However, this should not reduce code quality. Any other big project won't just accept users' PR and would require passing a code review. 

>

tommyettinger:
>  Working around the issue by using negative number looks like a dirty hack.

[Look who else uses the same "dirty hack."](https://github.com/openjdk/jdk/blob/ecec611af6c6314d7a834392f38468ad3f390e2d/src/java.base/share/classes/java/lang/Integer.java#L152-L202)

> Since there's a way to support concept of "unsigned integer" types using these xyzUnsigned methods, it looks more natural to use them, in order to reduce mental load on a person who reads the code. Another benefit from using unsigned methods is the possibility to avoid `isMin` flag.

You seem not to acknowledge that the xyzUnsigned methods are more, not less, mental load for some people other than yourself. Every technique in software is familiar to some people and unfamiliar to others. How would you feel if you were contributing to a Java repo and someone said, "No, your code isn't wrapping all side effects in a monad, we can't accept it." What if Java 21 introduces an optional side effect monad, and you suddenly need to refactor all of your code to include it, just because the language now includes this brand new feature? You CAN and SHOULD use necessary tools, right?

> > won't be recommending TeaVM for cross-platform apps until the library is more compatible.
> 
> Can you please explain what you mean?

I don't know of any way to, in my own code, patch in support for APIs that are missing or non-functional in TeaVM when I encounter them, e.g. Formatter support for floats and doubles. In GWT, it's relatively easy to super-source in replacement classes for the JDK or other libraries, or to switch out an implementation for one that uses JS-specific code. Importantly, super-sourcing in a library doesn't require that library to have a dependency on GWT libraries to function. I haven't seen any equivalent in TeaVM, so I don't know how a cross-platform library that has specialized behavior on the web could function. I'm limited to only what the TeaVM standard library has implemented right now if I want to use super-sourcing for GWT and the normal Java code path for anything else...

> > Consider being more flexible in the future.
> 
> Sorry, but I'm concerned about code quality here. For me it was not hard to fix the issue myself, I just want to help people to become contributors. However, this should not reduce code quality. Any other big project won't just accept users' PR and would require passing a code review.

Oh, so you fixed it! How'd the code review go? Did you have to explain how there had been a test for the max value, but not the min value? Whoever wrote that test probably wasn't concerned about code quality right then!

konsoletyper:
> I haven't seen any equivalent in TeaVM, so I don't know how a cross-platform library that has specialized behavior on the web could function. I'm limited to only what the TeaVM standard library has implemented right now if I want to use super-sourcing for GWT and the normal Java code path for anything else...

So, the alternative to "super sourcing" in GWT is the only show stopper for you? Then I would think on the ability to implement it. However, there's already ways to deal with what you mean:

1. `@PlatformMarker` annotation. See examples in the source code. If you are interested, I'll document it.
2. Using inversion of control pattern and introducing separate entry points for TeaVM and JVM implementations.
3. Creating three modules for platform-specific code: 'stub', 'jvm' and 'teavm', where stub consists of `native` methods or methods that throw some exception immediately. Then user logic module would depend on 'stub' (with `compileOnly` configuration in Gradle or `provided` scope in Maven), but actual entry points for TeaVM and JVM would depend on actual implementation methods.
4. Writing a TeaVM plugin which would transform IR and patch necessary parts.

And so on. Also, I would be glad to receive actual list of APIs that you need and I'll consider implementing them.

> Whoever wrote that test probably wasn't concerned about code quality right then!

"Whoever" was me almost 10 years ago. This does not mean that I'm was not concerned about code quality right then, this only means that I can make mistakes and that my experience since then has grown.

PR #684, opened 2023-04-04
Java reflection: return empty array when no reflection data available

danko-david:
getDeclaredMethods() and getDeclaredFields() call fails when no reflection data available. In this case process and return an empty array.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TClass.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TClass.java
@@ -45,6 +45,7 @@ import org.teavm.interop.DelegateTo;
 import org.teavm.interop.NoSideEffects;
 import org.teavm.interop.Unmanaged;
 import org.teavm.jso.core.JSArray;
+import org.teavm.jso.core.JSObjects;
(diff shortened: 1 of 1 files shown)


konsoletyper:
This should not work this way.

First, set of classes that are ever accessed via `getDeclaredMethods`/`getDeclaredFields`, is automatically computed ([here](https://github.com/konsoletyper/teavm/blob/3e1edeb90c561b5e489baed886efe1a14d139ded/classlib/src/main/java/org/teavm/classlib/impl/ReflectionDependencyListener.java#L115), so there should not be the case when 'reflection is not available'. Metadata is always available. The mechanism described in [documentation](https://teavm.org/docs/runtime/java-classes.html) only affects the ability to call method. If you encounter case when you access `getDeclaredMethods` or `getDeclaredFields` and get weird exception, this is affected by bug in detection of set of classes with metadata available. Can you please provide an example that reproduces this issue?

Second, if at some point a new mechanism, that allows to tune the set of classes with metadata, is introduced, I believe, in this case attempt to call `getDeclaredMethods` or `getDeclaredFields` should throw `SecurityException`.

PR #691, opened 2023-04-18
Record class name tostring

Ihromant:
Added class name to string. I'm not sure that it's the best way to determine simple name of class. Also test behaves strangely, it fails either in java or in js depending what assertEquals I'm using. Would be glad for advice how to fix it.


core/src/main/java/org/teavm/model/ValueType.java
```diff
@@ -37,6 +37,11 @@ public String getClassName() {
             return className;
         }
 
+        public String getSimpleName() {
```
konsoletyper, line 40: Please, remote this method.

classlib/src/main/java/org/teavm/classlib/impl/record/ObjectMethodsSubstitutor.java
```diff
@@ -156,7 +156,9 @@ private ValueEmitter hash(ProgramEmitter pe, ValueType type, ValueEmitter a) {
     private ValueEmitter substituteToString(DynamicCallSite callSite, ProgramEmitter pe) {
         ValueEmitter thisVar = callSite.getArguments().get(0);
         String names = callSite.getBootstrapArguments().get(1).getString();
-        ValueEmitter resultVar = pe.construct(StringBuilder.class, pe.constant("["));
+        ValueEmitter resultVar = pe.construct(StringBuilder.class, pe.constant(((ValueType.Object) thisVar.getType()).getSimpleName()));
```
konsoletyper, line 159: To properly get simple name, you should first acquire ClassReaderSource from DynamicCallSite, then use it to find ClassReader by name and get its simple name. If it's null, try to extract it from fully qualified name using current approach.

konsoletyper (changes requested):

konsoletyper:
>  Also test behaves strangely, it fails either in java or in js depending what assertEquals I'm using

What do you mean?

tests/src/test/java/org/teavm/vm/RecordTest.java
```diff
@@ -44,6 +44,12 @@ public void toStringMethod() {
 
         int index = 0;
 
+        // assertEquals("B[x=2, y=q, z=3]", s);
```
konsoletyper, line 47: I could not find any spec regarding Record.toString implementation, so I guess it may vary between JDK implementations and versions. This makes it impossible to test toString implementation this way.

PR #698, opened 2023-05-12
Integer long bit fixes

Ihromant:
Fixed Integer.highestOneBit for negative values (tests added). Long version is working, but changed it for consistency. Replaced lowestOneBit to more performant versions.


Ihromant:
@konsoletyper something is wrong with int unsigned shift in corner case. Added test with commented case which is successful in Java, but fails in JS.

classlib/src/main/java/org/teavm/classlib/java/lang/TInteger.java
```diff
@@ -316,11 +316,11 @@ public static int numberOfTrailingZeros(int i) {
     }
 
     public static int highestOneBit(int i) {
-        return 0x80000000 >>> numberOfLeadingZeros(i);
+        return i & (0x80000000 >>> numberOfLeadingZeros(i));
```
konsoletyper, line 319: Not sure why this `i &` necessary here. The left operand for shift is still one bit.

tests/src/test/java/org/teavm/classlib/java/lang/IntegerTest.java
```diff
@@ -26,6 +26,13 @@
 @RunWith(TeaVMTestRunner.class)
 @WholeClassCompilation
 public class IntegerTest {
+    @Test
+    public void testRightUnsignedShift() {
+        assertEquals(1 << 31, Integer.MIN_VALUE >>> 0);
+        assertEquals(0, Integer.numberOfLeadingZeros(-1));
+        // TODO this test fails assertEquals(1 << 31, Integer.MIN_VALUE >>> Integer.numberOfLeadingZeros(-1));
```
konsoletyper, line 33: I fixed the issue, now you can uncomment this

classlib/src/main/java/org/teavm/classlib/java/lang/TInteger.java
Ihromant: This is needed for case i == 0. Added test which covers this case.

PR #699, opened 2023-05-15
Collectors grouping by

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
```diff
@@ -170,6 +170,14 @@ default void forEach(BiConsumer<? super K, ? super V> action) {
         }
     }
 
+    default void replaceAll(BiFunction<? super K, ? super V, ? extends V> function) {
+        TIterator<Entry<K, V>> iterator = entrySet().iterator();
```
konsoletyper, line 174: I would use `var` here

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
```diff
@@ -170,6 +170,14 @@ default void forEach(BiConsumer<? super K, ? super V> action) {
...
+        while (iterator.hasNext()) {
+            TMap.Entry<K, V> next = iterator.next();
```
konsoletyper, line 176: I would use `var` here

PR #708, opened 2023-06-03
Added the copyOf method to List, Set and Map interfaces

siepkes:
The `copyOf` static method was added in Java 10 to the `List`, `Set` and `Map` interfaces. 

Since no other additions were made since Java 10 to these interfaces this commit brings the `List`, `Set` and `Map` interfaces to 100% completion for the latest LTS (Java 17) at the time of writing.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
@@ -294,4 +294,8 @@ public interface TMap<K, V> {
     static <K, V> TMap.Entry<K, V> entry(K k, V v) {
         return new TTemplateCollections.ImmutableEntry<>(requireNonNull(k), requireNonNull(v));
     }
+
+    static <K, V> TMap<K, V> copyOf(TMap<? extends K, ? extends V> map) {
+        return new TTemplateCollections.NEtriesMap<>(map);
+    }
 }
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TTemplateCollections.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TTemplateCollections.java
@@ -36,6 +36,23 @@ public final class TTemplateCollections {
             this.list = list;
         }
 
+        @SuppressWarnings("unchecked")
+        public ImmutableArrayList(TCollection<? extends T> collection) {
+            T[] list = (T[]) new Object[collection.size()];
+
+            TIterator<? extends T> iter = collection.iterator();
+            int index = 0;
+            while (iter.hasNext()) {
+                T element = iter.next();
+                if (element == null) {
+                    throw new NullPointerException();
+                }
+                list[index++] = element;
+            }
+
+            this.list = list;
+        }
+
         @Override
         public T get(int index) {
             return list[index];
@@ -193,6 +210,9 @@ public final class TTemplateCollections {
     static class NElementSet<T> extends AbstractImmutableSet<T> {
         private T[] data;
 
+        /**
+         * Throws an exception on duplicate elements.
+         */
         @SafeVarargs
         NElementSet(T... data) {
             T[] table = data.clone();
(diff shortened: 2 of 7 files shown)


classlib/src/main/java/org/teavm/classlib/java/util/TTemplateCollections.java
```diff
@@ -416,17 +472,31 @@ public TCollection<V> values() {
...
+            this.data = toEntryArray(map.entrySet().toArray(new TMap.Entry[0]));
+        }
+
+        /**
+         * Creates an array where the {@code Entry} elements are positioned in such a way they
+         * are compatible with the contract of {@link java.util.Map}.
+         */
+        @SuppressWarnings("unchecked")
+        private Entry<K, V>[] toEntryArray(Entry<K, V>[] entries) {
+            Entry<K, V>[] table = new Entry[entries.length];
```
konsoletyper, line 489: Does this line pass checkstyle?

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
```diff
@@ -294,4 +294,8 @@ static <K, V> TMap<K, V> ofEntries(TMap.Entry<K, V>... entries) {
     static <K, V> TMap.Entry<K, V> entry(K k, V v) {
         return new TTemplateCollections.ImmutableEntry<>(requireNonNull(k), requireNonNull(v));
     }
+
+    static <K, V> TMap<K, V> copyOf(TMap<? extends K, ? extends V> map) {
```
konsoletyper, line 298: I believe `NEntriesMap` can override this map to simply return itself without copying.

konsoletyper (changes requested):

classlib/src/main/java/org/teavm/classlib/java/util/TTemplateCollections.java
siepkes: It does, I wouldn't dare to submit code which doesn't pass tests or checkstyle ;-). Is there something you want me to modify here? Maybe I'm missing the obvious but I don't think I see any style problems with it.

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
siepkes: I don't think I fully understand. The API of `copyOf()` specifies: 

> Returns an unmodifiable Map containing the entries of the given Map. The given Map must not be null, and it must not contain any null keys or values. If the given Map is subsequently modified, the returned Map will not reflect such modifications.

I interpreted that as: If for example a `HashMap` is passed the `copyOf` method must return a new immutable map. 

I could add a check to see if the passed argument is already an (unmodifiable) `NEntriesMap` instance. In which case we can simply return that. Is that what you meant?

konsoletyper: > Is that what you meant?

Yes, that's what I mean. However, you don't need to check anything with `instanceof`, let polymorphism work for you. Just write specific implementation in an immutable implementation.

siepkes: Thanks for the clarification! However I don't think using polymorphism will work for this particular case. As far as I know (granted, that's a low bar ;-) this can in practice only be implemented with `instanceof`. For example adding this method to `Maps`:

```java
static <K, V> TMap<K, V> copyOf(TTemplateCollections.NEtriesMap<K, V> map) {
    return map;
}
```

and then using a test such as below will fail:

```java
@Test
public void copyOfOptimized() {
    Map<String, Integer> map = Map.of("q", 0, "w", 1, "e", 2);
    Map<String, Integer> mapCopy1 = Map.copyOf(map);
    Map<String, Integer> mapCopy2 = Map.copyOf(mapCopy1);

    assertSame("Must not create copies of immutable collections", mapCopy1, mapCopy2);
}
```

Because the method overloading will only select the `copyOf(TTemplateCollections.NEtriesMap<K, V> map)` method if the caller explicitly calls the method with a `NEtriesMap` type. However in practice this won't happen since the caller should never be aware they are using the `NEtriesMap` implementation of the `Map` interface. 

Changing the `copyOf` implementation to something like this does pass the test:

```java
@SuppressWarnings("unchecked")
static <K, V> TMap<K, V> copyOf(TMap<? extends K, ? extends V> map) {
    if (map instanceof TTemplateCollections.NEtriesMap) {
        return (TTemplateCollections.NEtriesMap<K, V>) map;
    } else {
        return new TTemplateCollections.NEtriesMap<>(map);
    }
}
```

konsoletyper: Ah, did not notice `static` modifier. Then you are right, there's no way to apply polymorphism here.

classlib/src/main/java/org/teavm/classlib/java/util/TTemplateCollections.java
siepkes: I think the right hand side (`new Entry[]`) is seen as a unchecked cast but not as raw usage of parameterized class. The left hand side (`Entry<K, V>`) would be seen as raw usage of parameterized class if it was without type definition (so `Entry[] table` instead of `Entry<K, V>[] table`). In that case it would indeed likely trigger a checkstyle raw usage error. 

I thought about it a bit but I also don't really see an alternative way of doing it (with better generics) compared to the current implementation.

classlib/src/main/java/org/teavm/classlib/java/util/TMap.java
siepkes: I've updated the PR to prevent it from copying `NEtriesMap` instances. I've also added a test for this to ensure this behavior doesn't silently break in the future.

Issue #713, opened 2023-06-18
ByteBuffer,CharBuffer,*Buffer throws IOOB unexpectedly


kirillp:
The following code works on OpenJDK but throws unexpected IndexOutOfBoundsException on TeaVM

```
  public static void test() {
    ByteBuffer bb = ByteBuffer.allocate(0);
    bb.put(new byte[0]);
    bb.get(new byte[0]);

    CharBuffer cb = CharBuffer.allocate(0);
    cb.put("");
    cb.get(new char[0]);

    ShortBuffer sb = ShortBuffer.allocate(0);
    sb.put(new short[0]);
    sb.get(new short[0]);

    IntBuffer ib = IntBuffer.allocate(0);
    ib.put(new int[0]);
    ib.get(new int[0]);

    FloatBuffer fb = FloatBuffer.allocate(0);
    fb.put(new float[0]);
    fb.get(new float[0]);

    DoubleBuffer db = DoubleBuffer.allocate(0);
    db.put(new double[0]);
    db.get(new double[0]);

    LongBuffer lb = LongBuffer.allocate(0);
    lb.put(new long[0]);
    lb.get(new long[0]);
  }

```

PR #714, opened 2023-06-18
fix for #713

kirillp:
fix ByteBuffer, CharBuffer, *Buffer IOOB exception throw with zero length arrays in arguments (#713)


classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
```diff
@@ -425,13 +425,8 @@ public TString concat(TString str) {
...
 
         var buffer = new char[length() + str.length()];
-        int index = 0;
-        for (int i = 0; i < length(); ++i) {
-            buffer[index++] = charAt(i);
-        }
-        for (int i = 0; i < str.length(); ++i) {
-            buffer[index++] = str.charAt(i);
-        }
+        System.arraycopy(characters, 0, buffer, 0, characters.length);
```
konsoletyper, line 428: Please, roll back this change

PR #716, opened 2023-06-20
Collectors partitioning by

Ihromant:


konsoletyper:
I usually accept such PRs, since they are correct, but I would like to encourage you again to start using `var` keyword when possible.

PR #741, opened 2023-09-18
Float/double equals fixes

Ihromant:
Please consider fix for (supposedly) rt_compare, which will fix failing in JS env test.
https://github.com/konsoletyper/teavm/compare/master...Ihromant:teavm:double-equals-fixes?expand=1#diff-82a57ed5fe3b9c9c8e7ff9af89119d3faf3fe8ba2a4519eaec3c203411540c09R137

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TFloat.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TFloat.java
@@ -31,6 +31,7 @@ public class TFloat extends TNumber implements TComparable<TFloat> {
     public static final int MAX_EXPONENT = 127;
     public static final int MIN_EXPONENT = -126;
     public static final int SIZE = 32;
+    public static final int BYTES = SIZE / Byte.SIZE;
     public static final Class<Float> TYPE = float.class;
     private float value;
 
@@ -84,7 +85,7 @@ public class TFloat extends TNumber implements TComparable<TFloat> {
         if (this == other) {
             return true;
         }
-        return other instanceof TFloat && ((TFloat) other).value == value;
+        return other instanceof TFloat && floatToIntBits(((TFloat) other).value) == floatToIntBits(value);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


Ihromant:
First issue introduced in this commit is that $rt_compare fails when we are putting 0.0 and -0.0. It returns 0 when it needs to be 1. In java side this behavior is correct.

Ihromant:
I don't touch this function because it seems that it is used alot in compilations/translations.

classlib/src/main/java/org/teavm/classlib/java/lang/TFloat.java
```diff
@@ -84,7 +85,7 @@ public boolean equals(Object other) {
         if (this == other) {
             return true;
         }
-        return other instanceof TFloat && ((TFloat) other).value == value;
+        return other instanceof TFloat && floatToIntBits(((TFloat) other).value) == floatToIntBits(value);
```
konsoletyper, line 88: I'm not sure if it has good approach WRT performance. Also, I'd check various cases with infinity and NaN.

konsoletyper (changes requested):

classlib/src/main/java/org/teavm/classlib/java/lang/TFloat.java
Ihromant: This is how equals works in JDK. It compares doubles bitwise, not with == operator.
So, Float.NaN != Float.NaN, but Float.valueOf(Float.NaN).equals(Float.NaN).
I don't know why this is done that way, but I think that main reason for this is to make equality relation reflexive and therefore equivalence relation (which would not be in case of comparing by == for NaN).
I'm not insisting on this change, I never do Map<Double, Something>, but this is how it works in JDK.
Added test cases which cover this behavior (will be failing when rollback this change).

Ihromant: By the way, hashCode is done same way: floatToIntBits(value).

Ihromant: One more point to follow consistency.

konsoletyper:
Plese, revert 'Fixed comparator in runtime.js'

Ihromant:
I get it from posted question on codegolf stackexchange. https://codegolf.stackexchange.com/questions/265365/comparator-function-in-javascript#265366
Strangely, but it works and covers all cases in tests. If you wish so, I will revert it, but it seems to be performant enough, simple and correct.

konsoletyper:
Please, revert. If you want this particular change, we can discuss it later in a separate PR.

PR #743, opened 2023-09-18
RandomGenerator

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/util/random/TRandomGenerator.java
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+
+    private static void checkRange(long origin, long bound) {
+        if (origin >= bound) {
+            throw new IllegalArgumentException();
+        }
+    }
+
+    static double[] pairGaussian(TRandomGenerator rng) {
```
konsoletyper, line 279: I'd moved this method into separate class  somewhere in `org.teavm.classlib.impl` package.

konsoletyper (changes requested):

PR #744, opened 2023-09-18
Fix commons-cli not being vendored

SquidDev:
The relocating of `org.apache.commons` also includes `org.apache.commons.cli`. This means attempting to run the CLI fails as it cannot find classes under `org.teavm.apachecommons.cli`.

Vendoring commons-cli as well as commons-io fixes this issue.


konsoletyper:
@SquidDev BTW, did you know that TeaVM now provides its own Gradle plugin, so you don't need to run JavaExec anymore

PR #745, opened 2023-09-19
Fixed comparator in runtime.js

Ihromant:
Comparator function obtained by beautiful answer here https://codegolf.stackexchange.com/questions/265365/comparator-function-in-javascript#265366

### DIFF
--- a/core/src/main/resources/org/teavm/backend/javascript/runtime.js
+++ b/core/src/main/resources/org/teavm/backend/javascript/runtime.js
@@ -24,7 +24,8 @@ function $rt_nextId() {
     return x;
 }
 function $rt_compare(a, b) {
-    return a > b ? 1 : a < b ? -1 : a === b ? 0 : 1;
+    var x;
+    return (x = (a > b) - (b > a)) ? x : (1/a > 1/b) - (1/b > 1/a) + (b == b) - (a == a);
 }
 function $rt_isInstance(obj, cls) {
     return obj instanceof $rt_objcls() && !!obj.constructor.$meta && $rt_isAssignable(obj.constructor, cls);
--- a/core/src/main/resources/org/teavm/backend/javascript/runtime.js
+++ b/core/src/main/resources/org/teavm/backend/javascript/runtime.js
@@ -24,8 +24,8 @@ function $rt_nextId() {
     return x;
 }
 function $rt_compare(a, b) {
-    var x;
-    return (x = (a > b) - (b > a)) ? x : (1/a > 1/b) - (1/b > 1/a) + (b == b) - (a == a);
+    var diff = (a > b) - (b > a);
+    return diff ? diff : (1/a > 1/b) - (1/b > 1/a) + (b == b) - (a == a);
 }
 function $rt_isInstance(obj, cls) {
     return obj instanceof $rt_objcls() && !!obj.constructor.$meta && $rt_isAssignable(obj.constructor, cls);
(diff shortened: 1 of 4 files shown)


konsoletyper:
I used to have such "fix" which I reverted couple days later. Don't remember why, though. I'd run all tests and I'd also tried couple of my apps with this fix. Another concern is: this `$rt_compare` is used both for int and float/double typed, which doesn't seem correct. Thus it should be split for these cases, and `int` case should remain simple and fast. And this in turn must be taken into account in compiler/optimizer case.

Ihromant:
It works for ints as well because ints are the same "numbers" in JS. Still, it can be simplified for performance reasons.

konsoletyper:
It won't work well in all cases, because there aren't '+0' and '-0' for ints, but there are for JS numbers. Since I'm not sure if int is properly emulated everywhere, `$rt_compare` for ints should take this into account. Or rather compiler should be updated to handle such cases.

Ihromant:
Could you please consider following case (extracted it from already merged PR) in compiler/optimizer?

Let's consider java code for signum:
```java
public static double signum(double d) {
        if (d == 0.0 || Double.isNaN(d)) {
            return d;
        }
        return d < 0.0 ? -1.0 : 1.0;
    }
```
For some reason (optimizations etc.) it converts to code which is not the same as java code:
```javascript
function jl_Math_signum($d) {
        var var$2;
        var$2 = $rt_compare($d, 0.0);
        if (var$2 && !($rt_globals.isNaN($d) ? 1 : 0))
            return var$2 >= 0 ? 1.0 : (-1.0);
        return $d;
    }
```
Main issue of compiled code is that it replaces $d==0.0 to $rt_compare result (or Double.compare in Java) which is wrong in case of passing -0.0 to this function.

This could lead to strange bugs.

Ihromant:
Added tests to show that -0 and 0 are handled properly in int case.

core/src/main/resources/org/teavm/backend/javascript/runtime.js
```diff
@@ -24,7 +24,8 @@ function $rt_nextId() {
     return x;
 }
 function $rt_compare(a, b) {
-    return a > b ? 1 : a < b ? -1 : a === b ? 0 : 1;
+    var diff = (a > b) - (b > a);
```
Ihromant, line 27: This diff can be extracted as rt_compare for int.

konsoletyper:
> For some reason (optimizations etc.) it converts to code which is not the same as java code:

The produced code is correct. See bytecode generated by javac:

```
ublic static double signum(double);
    Code:
       0: dload_0
       1: dconst_0
       2: dcmpl
       3: ifeq          13
       6: dload_0
       7: invokestatic  #7                  // Method java/lang/Double.isNaN:(D)Z
      10: ifeq          15
      13: dload_0
      14: dreturn
      15: dload_0
      16: dconst_0
      17: dcmpg
      18: ifge          27
      21: ldc2_w        #13                 // double -1.0d
      24: goto          28
      27: dconst_1
      28: dreturn
```

And see spec for [dcmpl](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.dcmp_op) instruction.

So implementation of `$rt_compare`, which is used to emulate this instruction is correct and should not be changed.

Ihromant:
Got it, thanks. At least, in this case Double.compare should use other method (fixed one for instance), but not rt_compare (which obviously is not equal to Double.compare).

Ihromant:
Also if it appears in bytecode dcmpl, then it should be replaced with
```javascript
function $rt_compare_l(a, b) {
    return a > b ? 1 : a < b ? -1 : a === b ? 0 : -1;
```
(still, I don't know how often dcmpl appears in bytecode

Ihromant:
@konsoletyper 
created two separate issues which I don't want to fix because I can break something:
https://github.com/konsoletyper/teavm/issues/747
https://github.com/konsoletyper/teavm/issues/748
and closing this PR because it's incorrect, but tests can be reused when second issue is fixed.

PR #751, opened 2023-09-21
classlib: Add nio's channel interfaces

SquidDev:
- Add the basic `java.nio.channels.Channel` interfaces.
 - Add several nio exception types.
 - Add `UncheckedIOException`

Afraid I've only added the types that my existing code needs (though the interface for each should be complete). Happy to flesh this out further if so desired!

Interfaces for each was based on the Javadoc, aside from `FileSystemException` which was based on observed behaviour - can confirm I've not looked at the JDK sources here.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/io/TUncheckedIOException.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/io/TUncheckedIOException.java
@@ -1,0 +1,15 @@
+package org.teavm.classlib.java.io;
+
+import java.io.IOException;
+
+public class TUncheckedIOException extends RuntimeException {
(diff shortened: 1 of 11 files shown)


konsoletyper:
@SquidDev Please, next time you create a PR make sure you build your working copy with "build" run configuration in IDEA. If you don't have IDEA, you can use following command:

```
./gradlew build -x test --stacktrace
```

SquidDev:
Ahh, I am so sorry! I had done an `./gradlew assemble`, but forgotten about checkstyle. Would you like me to submit a PR to fix those?

konsoletyper:
Already fixed that

SquidDev:
Fantastic, thank you! Many apologies for that.

PR #752, opened 2023-09-21
Fixes for hashcode and other minor fixes/additions

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/lang/TBoolean.java
```diff
@@ -97,11 +88,11 @@ public static boolean getBoolean(String key) {
     }
 
     public static boolean logicalAnd(boolean a, boolean b) {
-        return a & b;
+        return a && b;
```
konsoletyper, line 91: What's the purpose of this change?

classlib/src/main/java/org/teavm/classlib/java/lang/TInteger.java
```diff
@@ -45,7 +45,7 @@ public static String toString(int i, int radix) {
     }
 
     public static int hashCode(int value) {
-        return (value >>> 4) ^ (value << 28) ^ (value << 8) ^ (value >>> 24);
+        return value;
```
konsoletyper, line 48: What's the purpose of this change?

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
```diff
@@ -1463,10 +1458,9 @@ public static int hashCode(float[] a) {
         if (a == null) {
             return 0;
         }
-        int hash = 0xA5A537FC;
+        int hash = 1;
         for (int i = 0; i < a.length; ++i) {
-            int h = TFloat.floatToIntBits(a[i]);
-            hash = TInteger.rotateLeft(h, 4) ^ TInteger.rotateRight(h, 7) ^ TInteger.rotateLeft(hash, 13);
+            hash = 31 * hash + Float.hashCode(a[i]);
```
konsoletyper, line 1463: What's the purpose of changes like this?

konsoletyper (changes requested):

classlib/src/main/java/org/teavm/classlib/java/lang/TInteger.java
Ihromant: Performance improvement.

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
Ihromant: If you check the documentation of Arrays, you will see that Arrays.hashCode should be equal to hashcode of equal list.

classlib/src/main/java/org/teavm/classlib/java/lang/TBoolean.java
Ihromant: Correspondence of name to the implementation.

classlib/src/main/java/org/teavm/classlib/java/lang/TInteger.java
konsoletyper: This *is* a performance improvement for running this particular function, but will it make object distribution in hashtable better? Do you have any CS papers with overview of quality of hashing functions? I'm no competent enough, but naively it seems that applying at least *some* transformation to value is better than not doing so.

Ihromant: And also documentation states that hashcode should be equal to primitive int, not this strange permutation.

classlib/src/main/java/org/teavm/classlib/java/lang/TBoolean.java
konsoletyper: But in this particular context what's the difference between & and &&?

classlib/src/main/java/org/teavm/classlib/java/lang/TInteger.java
konsoletyper: Ah, I see. I wonder why Java guys added this requirement. In previous versions of Java particular implementation of Integer.hashCode was not fixed.

classlib/src/main/java/org/teavm/classlib/java/lang/TBoolean.java
Ihromant: No difference but short-circuiting. Actually this can be reverted if needed.

konsoletyper (changes requested):

PR #755, opened 2023-09-22
Add Java 9's Arrays.equals and Arrays.mismatch

SquidDev:
This adds support for Java 9's ranged `Arrays.equals`, as well as the new `Arrays.mismatch` function.

The implementation is the same for every primitive, so hopefully this isn't as terrible to review as it looks. The duplication is a shame, but short of templating the file, I'm not sure there's a better solution here.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
@@ -1253,6 +1253,25 @@ public class TArrays extends TObject {
         }
     }
 
+    private static int mismatchImpl(long[] a, long[] a2, int length) {
+        for (int i = 0; i < length; ++i) {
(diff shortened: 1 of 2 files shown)


classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
```diff
@@ -1253,19 +1253,81 @@ public static <T> int binarySearch(T[] a, int fromIndex, int toIndex, T key, TCo
         }
     }
 
+    private static int mismatchImpl(long[] a, long[] a2, int length) {
```
konsoletyper, line 1256: I'd used one version of `misatchImpl` for both cases, this is an excessive one.

konsoletyper (changes requested):

PR #759, opened 2023-09-24
Added SequencedCollection interface, inherited List from it

Ihromant:
Added SequencedCollection interface, inherited List from it (initial part of JEP-431).

Can't properly cover it by tests (JDK-21 is not build JDK for teavm now), but it should work according to the common logic. Not reimplemented ListIterator and Iterator for reversed(), so they should not work properly for now (need tests to do it). Other reversed() overridden methods (and some not overridden) should work correctly because they have obvious implementation.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TList.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TList.java
@@ -16,6 +16,7 @@
 package org.teavm.classlib.java.util;
 
 import java.util.Objects;
+import java.util.RandomAccess;
 import org.teavm.classlib.java.util.function.TUnaryOperator;
 
 public interface TList<E> extends TSequencedCollection<E> {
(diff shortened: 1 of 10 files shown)


konsoletyper:
> Can't properly cover it by tests (JDK-21 is not build JDK for teavm now)

But what's the problem to switch?

> Not reimplemented ListIterator and Iterator for reversed(), so they should not work properly for now (need tests to do it). 

Default implementations of `iterator`/`listIterator` in `AbstractList` utilize `get` and `size`, so whenever you implemented them properly, they also should work.

classlib/src/main/java/org/teavm/classlib/java/util/TList.java
```diff
@@ -50,6 +50,94 @@ default void sort(TComparator<? super E> c) {
...
+    default E removeLast() {
+        if (isEmpty()) {
+            throw new TNoSuchElementException();
+        }
+        return remove(this.size() - 1);
+    }
+
+    @Override
+    default TList<E> reversed() {
+        return new TAbstractList<>() {
```
konsoletyper, line 97: Other methods worth implementing here (only for performance reasons): addAll, clear, isEmpty, contains, addFirst, addLast, getFirst, getLast, removeFirst, removeLast, reversed, listIterator

konsoletyper:
It's useful to implement these `*First`/`*Last` methods in `LinkedList`

Ihromant:
@konsoletyper I suppose that implementation is finished, please review.

classlib/src/main/java/org/teavm/classlib/java/util/TList.java
```diff
@@ -159,4 +208,78 @@ static <E> TList<E> of(E... elements) {
     static <E> TList<E> copyOf(TCollection<? extends E> collection) {
         return new TTemplateCollections.ImmutableArrayList<>(collection);
     }
+
+    class ReversedList<E> extends TAbstractList<E> {
```
konsoletyper, line 212: I'd brought it to top level. And made it package-private

konsoletyper: Implementing `iterator` would also be quite useful here

classlib/src/main/java/org/teavm/classlib/java/util/TReversedList.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.classlib.java.util;
+
+class TReversedList<E> extends TAbstractList<E> {
```
konsoletyper, line 18: getFirst/getLast too, please

classlib/src/main/java/org/teavm/classlib/java/util/TList.java
Ihromant: It uses implementation of iterator from AbstractList which is simple and performant.

konsoletyper: It's not. Underlying list does not necessarily implement `get` in the most efficient ways. Sometimes it's much faster to iterate over list than getting elements with `get` method.

PR #762, opened 2023-09-25
Fixes for min and max

Ihromant:
Fix for Math.min and Math.max.
API note for JDK states that
```
API Note:
This method corresponds to the minimum/maximum operation defined in IEEE 754.
```
so I suppose that it could be converted to native call everywhere. Still, I don't know wasm and C, so added correct fallback for this platforms.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
@@ -206,7 +206,7 @@ public final class TMath extends TObject {
         if (a != a) {
             return a;
         }
-        if (a == 0.0 && b == 0.0d && 1 / b == Double.NEGATIVE_INFINITY) {
+        if (a == 0.0 && b == 0.0 && 1 / b == Double.NEGATIVE_INFINITY) {
             return b;
         }
         return a <= b ? a : b;
@@ -223,7 +223,7 @@ public final class TMath extends TObject {
         if (a != a) {
             return a;
         }
-        if (a == 0.0d && b == 0.0d && 1 / a == Double.NEGATIVE_INFINITY) {
+        if (a == 0.0 && b == 0.0 && 1 / a == Double.NEGATIVE_INFINITY) {
             return b;
         }
         return a >= b ? a : b;
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TMath.java
@@ -195,20 +195,72 @@ public final class TMath extends TObject {
         return a > b ? a : b;
     }
 
+    @GeneratedBy(MathNativeGenerator.class)
+    @NoSideEffects
+    private static native float minImpl(double a, double b);
+
     public static double min(double a, double b) {
-        return a < b ? a : b;
+        if (PlatformDetector.isJavaScript()) {
+            return minImpl(a, b);
+        }
+        if (a != a) {
+            return a;
+        }
+        if (a == 0.0 && b == 0.0d && 1 / b == Double.NEGATIVE_INFINITY) {
+            return b;
+        }
+        return a <= b ? a : b;
     }
+
+    @GeneratedBy(MathNativeGenerator.class)
+    @NoSideEffects
+    private static native float maxImpl(double a, double b);
 
     public static double max(double a, double b) {
-        return a > b ? a : b;
+        if (PlatformDetector.isJavaScript()) {
+            return maxImpl(a, b);
+        }
+        if (a != a) {
+            return a;
+        }
+        if (a == 0.0d && b == 0.0d && 1 / a == Double.NEGATIVE_INFINITY) {
+            return b;
+        }
+        return a >= b ? a : b;
     }
+
+    @GeneratedBy(MathNativeGenerator.class)
+    @NoSideEffects
+    private static native float minImpl(float a, float b);
 
     public static float min(float a, float b) {
-        return a < b ? a : b;
+        if (PlatformDetector.isJavaScript()) {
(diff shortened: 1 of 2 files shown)


konsoletyper:
What's the problem to learn how to use/test/patch Wasm and C backends?

konsoletyper:
BTW, here's what I found: https://2pi.dk/2016/05/ieee-min-max

Ihromant:
Not a problem, just need time and motivation to do this.
I have almost no experience with C/C++, so this is main issue that complicates my understanding.

Ihromant:
Regarding the link, it seems that it's even more complicated. But it seems that JS Math.min/max work similarly to Java Math.min/max.

konsoletyper:
> I have almost no experience with C/C++, so this is main issue that complicates my understanding.

Sure. Improving your skill can bring you better career opportunities. By contributing to OSS you can learn new things, like what you recently learned about IEEE 754. In this case you don't need to write from scratch, you need just add code to the existing one. You don't even need write a line of C code, you just need to find out how min/max functions called in C (hint: fmin, fmax) and apply corresponding annotation. 

As for Wasm, it's not based on C at all. Anyway, it's a bit complicated: you need to add intrinsic that turns min/max calls into corresponding Wasm instructions (namely, `WasmFloatBinary` with operation set to `MIN` and `MAX`). I suggest to add this functionality to `WasmRuntime`, for which there are already `WasmRuntimeIntrinsic` with examples of comparison functions. 

As for running tests, you need to set `teavm.tests.wasm` and `teavm.tests.c` to `true`. You can either edit (but don't commit!) `gradle.properties` file or pass following to command line: `-Pteavm.tests.wasm=true`

Ihromant:
I took a look where you pointed.
Frankly speaking, it's either some mess in implementation or I don't understand anything. When I add annotation
```java
@Import(module = "teavmMath", name = "min")
```
to both min(double a, double b) and min(float a, float b) i get "Function min already defined in this module" error.
When replacing it to 
```java
    public static float min(float a, float b) {
        return (float) min(a, (double) b);
    }
```
then wasm tests are running fine, but next issue is with C tests. I took a look at WasmCRenderingVisitor, looks like it should work, but it doesn't. Error is the following:
```
java.lang.RuntimeException: C compiler error
```
I can assume that I'm dumb, but...
In Math class there is pow function. And I can see no trace for this function implementation anywhere, but magically it works.
Shortly speaking, I don't know where to put magic to make it working.

konsoletyper:
> ```@Import(module = "teavmMath", name = "min")```

No, please, don't! I mentioned intrinsic. In Wasm there are instructions for min and max. Using JavaScript `Math` object is slower and isn't available in WASI.

> Function min already defined in this module

Strange, I'll investigate this.

> ```java.lang.RuntimeException: C compiler error```

Ah, yes. What operating system do you have? Do you have any C compiler installed?

Ihromant:
XUbuntu 22.04, I suppose C compiler is present by default. In my case it's gcc.
```
ihromant@ihromant-desktop:~$ gcc
gcc: fatal error: no input files
compilation terminated.
```
> First, you can always ask for help. Second, I explained where to put magic for this particular case.

Sorry for little bit emotional speech, generally it was ask for help with explained issues found and troubles.

konsoletyper:
> I think that issue is not related to C compiler, I wrote test with Math.pow and it worked everywhere.

Ok, but what did you write? And did console provide any additional output from C compiler?

konsoletyper:
Another hint for tests: did you know that they are actually compiled to `build/teavm/tests/(fully/quialified/path/to/Class)`? In case of JS there's JS file and number of html files to open JS. In case of C there are generated C files, which you can copile with following command:

```
gcc -g -O0 -lrt -lm all.c -o run_test
```

and then run

```
./run_test fully.qualified.ClassName.methodName()V
```

Ihromant:
```java
@GeneratedBy(MathNativeGenerator.class)
    @Import(module = "teavmMath", name = "min")
    @Unmanaged
    public static native double min(double a, double b);

    @GeneratedBy(MathNativeGenerator.class)
    @Import(module = "teavmMath", name = "max")
    @Unmanaged
    public static native double max(double a, double b);

    public static float min(float a, float b) {
        return (float) min(a, (double) b);
    }

    public static float max(float a, float b) {
        return (float) max(a, (double) b);
    }
```
and enabled C tests for teavm.

konsoletyper:
Sure. And if you find out how to get C compiler output, it'll say something about min and max functions not being declared (and then ld failing because of that). Because in C they are called `fmin` and `fmax`, I pointed it out in my previous comment.

Ihromant:
@konsoletyper I don't know how to resolve wasm and C implementations. They are using the same annotation @Import, so I suppose that different methods should be produced, one in WasmSupport, second raw. Also I'm assuming that WasmSupport is wrong becase it uses (?) Javascript, not native. But where is correct place I don't understand.
Also, I was not successful with C. When I'm using simple annotation like
```java
@Import(name = "fmax")
```
i get 
```
/home/ihromant/workspace/teavm/tests/build/teavm-tests/org/teavm/classlib/java/lang/MathTest/minMax/test/classes/java/lang/Math.c:31: undefined reference to `fmax'
```
errors. I tried to read C code, but for now it is beyond my understanding.

konsoletyper:
@Ihromant it turns out that C functions do not implement IEEE 754 standard. This produces unexpected result:

```c
#include <math.h>
#include <stdio.h>

int main() {
    float x = fminf((float) NAN, (float) INFINITY);
    printf("%f\n", x);
    return 0;
}
```

Ihromant:
In this case this PR is correct and can be merged.

konsoletyper:
Pushed from Git

Issue #774, opened 2023-10-01
loadClass signatures missing from ClassLoader


ldubost:
The following signatures are missing from ClassLoader implementation:

 public Class<?> loadClass(String name);
 public Class<?> loadClass(String name, boolean result);

konsoletyper:
I believe, they shuold not be supported at all.

ldubost:
The reason I added them, is that a library was making use of this (probably to load some optional extension), and this breaks the code I've been trying to run at compile time, while at runtime there are not issues. Having the two functions that return null would help in that case supporting more libraries. Now I understand that you might want to keep the classlib only for actually implemented code.

konsoletyper:
I think `ClassLoader.loadClass` can be implemented just like `Class.forName` is implemented. Also, optimization for `loadClass("constantString")` can be added just like there's one for `Class.forName`.

PR #775, opened 2023-10-01
loadClass signatures missing from ClassLoader #774

ldubost:
Placeholder implementation which would allow programs which use this functions to still compile.
The proposed placeholder implementation returns null.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TClassLoader.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TClassLoader.java
@@ -61,6 +61,14 @@ public abstract class TClassLoader extends TObject {
         return new ByteArrayInputStream(Base64Impl.decode(bytes));
     }
 
+    public Class<?> loadClass(String name) {
+        return null;
+    }
+
+    public Class<?> loadClass(String name, boolean result) {
+        return null;
+    }
+
     public static InputStream getSystemResourceAsStream(String name) {
         return getSystemClassLoader().getResourceAsStream(name);
     }


konsoletyper:
TeaVM is an ahead-of-time compiler that's incapable of dynamic capabilities of JVM like class loading, so it's unlikely that TeaVM ever obtains support of class loading. If you have a project that uses some missing methods like this, you can write a TeaVM plugin that patches your classes.

ldubost:
You are mentioning a TeaVM plugin to patch classes ? I tried to find documentation about this, but could not find it on the TeaVM documentation. Would there be a link to a documentation or an example ?

konsoletyper:
There's no documentation. There's an interface `TeaVMPlugin`. You just implement it and use ServiceLoader convention to register it. What you need it to implement and register `ClassHolderTransformer`. Use emit API (see `ProgramEmitter`) to generate bodies of new methods or take bodies from other methods and copy them into new bodies. For example, in you case it can look like this:

```java
public class ClassLoaderTransformer implements ClassHolderTransformer {
    @Override
    public void transformClass(ClassHolder cls, ClassHolderTransformerContext context) {
        if (cls.getName().equals(ClassLoader.class.getName())) {
            var method = new MethodHolder(new MethodDescriptor("loadClass", String.class, boolean.class, Class.class));
            var pe = ProgramEmitter.create(method, context.getHierarchy());
            pe.constantNull(Class.class).returnValue();
        }
    }
}
```

another example is:

```java
public class ClassLoaderTransformer implements ClassHolderTransformer {
    @Override
    public void transformClass(ClassHolder cls, ClassHolderTransformerContext context) {
        if (cls.getName().equals(ClassLoader.class.getName())) {
            var method = new MethodHolder(new MethodDescriptor("loadClass", String.class, boolean.class, Class.class));
            var templateMethod = context.getHierarchy().getClassSource().get(ClassLoaderTransformer.class.getName())
                    .getMethod(new MethodDescriptor("loadClassTemplate", String.class, boolean.class, Class.class));
            var programTemplate = templateMethod.getProgram();
            var program = ProgramUtils.copy(programTemplate);
            method.setProgram(program);
        }
    }

    @SuppressWarnings("unused")
    private Class<?> loadClassTemplate(String name, boolean initialize) {
        return null;
    }
}
```

There are plenty examples of plugins in TeaVM source code, actually, runtime is implemented as a set of plugins.

Issue #776, opened 2023-10-01
WeakHashMap implementation missing


ldubost:
WeakHashMap (TWeakHashMap.java) is missing in the classlib.

PR #777, opened 2023-10-01
WeakHashMap implementation missing #776

ldubost:
Implementation of TWeakHashMap by copying THashMap.
This implementation does not implement the "Weak" behavior of WeakHashMap.
However this would still be quite useful for any code which uses WeakHashMap


konsoletyper:
Sorry, I can't accept this. Fair WeakHashMap implementation required

ldubost:
Hi,

Do you have any suggestion of what would need to be done to implement the disposing of the HashMap ? Is there an similar case in another class already implemented ?

konsoletyper:
Sorry, don't understand your question

PR #810, opened 2023-10-08
Tree set map fixes

Ihromant:
Last part of JEP-431.

Unfortunately can't run check due to
```
* What went wrong:
Could not load the value of field `aliasSet` of `sun.nio.cs.UTF_8` bean found in field `charset` of `org.gradle.api.internal.resources.FileCollectionBackedTextResource` bean found in field `config` of task `:classlib:checkstyleMain` of type `org.gradle.api.plugins.quality.Checkstyle`.
> java.lang.reflect.InvocationTargetException (no error message)
```
error when launching `./gradlew check`
But tests for TreeSet and TreeMap passed from IDEA successfully.


tests/src/test/java/org/teavm/classlib/java/util/TreeSetTest.java
```diff
@@ -0,0 +1,178 @@
...
+        Iterator<String> it = set.iterator();
+        assertEquals("a", it.next());
+        it.remove();
+        assertEquals("b", it.next());
+        it.remove();
+        assertEquals("c", it.next());
+        it.remove();
+        assertEquals("d", it.next());
+        it.remove();
+//      TODO fixme  assertEquals("e", it.next());
```
konsoletyper, line 166: Just pushed the fix for this issue, you can try to merge and uncomment this

konsoletyper (changes requested):

tests/src/test/java/org/teavm/classlib/java/util/TreeSetTest.java
Ihromant: Works, thanks.

PR #813, opened 2023-10-10
Tree map fixes

Ihromant:
Added lots of various fixes for TreeMap.
Explanation. I get https://github.com/pulseenergy/harmony-treemap/blob/master/src/test/java/org/apache/harmony/java/util/TreeMapTest.java this file (Apache 2.0 licensed), renamed it to TreeMapTest1, adapted to our testcases and noticed multiple failures. Then I fixed them one by one.
So, now 2 options are possible:
1. Have two test files for TreeMap, one normal and one "verbose".
2. Remove this TreeMapTest1 file.
Personally I don't want to extract exact cases that are failing or merge this tests because this is stupid work IMHO.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TAbstractMap.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TAbstractMap.java
@@ -237,14 +237,18 @@ public abstract class TAbstractMap<K, V> extends TObject implements TMap<K, V> {
         if (size() != other.size()) {
             return false;
         }
-        for (TIterator<? extends TMap.Entry<K, V>> iter = entrySet().iterator(); iter.hasNext();) {
-            TMap.Entry<K, V> entry = iter.next();
-            if (!other.containsKey(entry.getKey())) {
-                return false;
+        try {
+            for (var it = entrySet().iterator(); it.hasNext(); ) {
+                TMap.Entry<K, V> entry = it.next();
+                if (!other.containsKey(entry.getKey())) {
+                    return false;
+                }
+                if (!TObjects.equals(entry.getValue(), other.get(entry.getKey()))) {
+                    return false;
+                }
             }
-            if (!TObjects.equals(entry.getValue(), other.get(entry.getKey()))) {
-                return false;
-            }
+        } catch (ClassCastException | NullPointerException e) {
+            return false;
         }
         return true;
     }
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TAbstractMap.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TAbstractMap.java
@@ -238,7 +238,7 @@ public abstract class TAbstractMap<K, V> extends TObject implements TMap<K, V> {
             return false;
         }
         try {
-            for (var it = entrySet().iterator(); it.hasNext(); ) {
+            for (var it = entrySet().iterator(); it.hasNext();) {
                 TMap.Entry<K, V> entry = it.next();
                 if (!other.containsKey(entry.getKey())) {
                     return false;
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TNavigableMap.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TNavigableMap.java
@@ -38,12 +38,16 @@ public interface TNavigableMap<K, V> extends TSortedMap<K, V> {
 
     K higherKey(K key);
 
+    @Override
     Entry<K, V> firstEntry();
 
+    @Override
     Entry<K, V> lastEntry();
 
+    @Override
     Entry<K, V> pollFirstEntry();
 
+    @Override
     Entry<K, V> pollLastEntry();
 
     TNavigableMap<K, V> descendingMap();
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
@@ -17,7 +17,6 @@ package org.teavm.classlib.java.util;
 
 import org.teavm.classlib.java.io.TSerializable;
 import org.teavm.classlib.java.lang.TCloneable;
-import org.teavm.classlib.java.lang.TIllegalArgumentException;
 
 public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TSerializable, TNavigableMap<K, V> {
     static class TreeNode<K, V> extends SimpleEntry<K, V> {
@@ -199,9 +198,11 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
 
     TreeNode<?, V> findExact(Object key) {
         TreeNode<K, V> node = root;
+        @SuppressWarnings("unchecked")
+        K k = (K) key;
+        comparator.compare(k, k);
         while (node != null) {
-            @SuppressWarnings("unchecked")
-            int cmp = comparator.compare((K) key, node.getKey());
+            int cmp = comparator.compare(k, node.getKey());
             if (cmp == 0) {
                 return node;
             } else if (cmp < 0) {
@@ -378,20 +379,17 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
 
     @Override
     public TSortedMap<K, V> subMap(K fromKey, K toKey) {
-        if (comparator.compare(fromKey, toKey) > 0) {
-            throw new TIllegalArgumentException();
-        }
-        return new MapView<>(this, fromKey, true, true, toKey, false, true, false);
+        return subMap(fromKey, true, toKey, false);
     }
 
     @Override
     public TNavigableMap<K, V> headMap(K toKey) {
-        return new MapView<>(this, null, true, false, toKey, false, true, false);
+        return headMap(toKey, false);
     }
 
     @Override
     public TNavigableMap<K, V> tailMap(K fromKey) {
-        return new MapView<>(this, fromKey, true, true, null, false, false, false);
+        return tailMap(fromKey, true);
     }
 
     @Override
@@ -565,16 +563,16 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
     }
 
     static class EntrySet<K, V> extends TAbstractSet<Entry<K, V>> implements TSequencedSet<Entry<K, V>> {
+        private final TTreeMap<K, V> owner;
+        private final K from;
+        private final boolean fromIncluded;
+        private final boolean fromChecked;
+        private final K to;
+        private final boolean toIncluded;
+        private final boolean toChecked;
+        private final boolean reverse;
         private int modCount = -1;
-        private TTreeMap<K, V> owner;
-        private K from;
-        private boolean fromIncluded;
-        private boolean fromChecked;
-        private K to;
-        private boolean toIncluded;
-        private boolean toChecked;
         private int cachedSize;
-        private boolean reverse;
 
          EntrySet(TTreeMap<K, V> owner, K from, boolean fromIncluded, boolean fromChecked,
                 K to, boolean toIncluded, boolean toChecked, boolean reverse) {
@@ -590,19 +588,19 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
 
         @Override
         public int size() {
-            int size = cachedSize;
             if (modCount != owner.modCount) {
                 modCount = owner.modCount;
-                size = owner.size();
+                int size = owner.size();
+                TreeNode<K, V>[] fromPath = null;
                 if (fromChecked) {
-                    TreeNode<K, V>[] path = fromIncluded ? owner.pathToNext(from, true)
+                    fromPath = fromIncluded ? owner.pathToNext(from, true)
                             : owner.pathToExactOrNext(from, true);
-                    for (TreeNode<K, V> node : path) {
+                    for (TreeNode<K, V> node : fromPath) {
                         if (node.left != null) {
                             size -= node.left.size;
                         }
                     }
-                    size -= path.length;
+                    size -= fromPath.length;
                 }
                 if (toChecked) {
                     TreeNode<K, V>[] path = toIncluded ? owner.pathToNext(to, false)
@@ -613,10 +611,14 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
                         }
                     }
                     size -= path.length;
+                    if (fromPath != null && fromPath.length > 0 && path.length > 0 && owner.comparator.compare(
+                            fromPath[fromPath.length - 1].getKey(), path[path.length - 1].getKey()) == 0) {
+                        size++;
+                    }
                 }
                 cachedSize = size;
             }
-            return size;
+            return cachedSize;
         }
 
         @Override
@@ -757,7 +759,7 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
                 throw new TConcurrentModificationException();
             }
             if (last == null) {
-                throw new TNoSuchElementException();
+                throw new IllegalStateException();
             }
             var newRoot = owner.deleteNode(owner.root, last.getKey());
             if (owner.root != newRoot) {
@@ -772,21 +774,34 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
     }
 
     static class MapView<K, V> extends TAbstractMap<K, V> implements TNavigableMap<K, V>, TSerializable {
-        private int modCount = -1;
-        private int cachedSize;
-        private TTreeMap<K, V> owner;
-        private K from;
-        private boolean fromIncluded;
-        private boolean fromChecked;
-        private K to;
-        private boolean toIncluded;
-        private boolean toChecked;
+        private final TTreeMap<K, V> owner;
+        private final K from;
+        private final boolean fromIncluded;
+        private final boolean fromChecked;
+        private final K to;
+        private final boolean toIncluded;
+        private final boolean toChecked;
+        private final boolean reverse;
         private EntrySet<K, V> entrySetCache;
-        private boolean reverse;
         private NavigableKeySet<K, V> cachedNavigableKeySet;
 
         MapView(TTreeMap<K, V> owner, K from, boolean fromIncluded, boolean fromChecked,
                 K to, boolean toIncluded, boolean toChecked, boolean reverse) {
+            check(owner, from, fromChecked, to, toChecked);
+            this.owner = owner;
+            this.from = from;
+            this.fromIncluded = fromIncluded;
+            this.fromChecked = fromChecked;
+            this.to = to;
+            this.toIncluded = toIncluded;
+            this.toChecked = toChecked;
+            this.reverse = reverse;
+            if (reverse) {
+                owner.ensureRevertedComparator();
+            }
+        }
+
+        private void check(TTreeMap<K, V> owner, K from, boolean fromChecked, K to, boolean toChecked) {
             if (fromChecked) {
                 if (toChecked) {
                     if (owner.comparator.compare(from, to) > 0) {
@@ -800,14 +815,6 @@ public class TTreeMap<K, V> extends TAbstractMap<K, V> implements TCloneable, TS
                     owner.comparator.compare(to, to);
                 }
             }
-            this.owner = owner;
-            this.from = from;
-            this.fromIncluded = fromIncluded;
-            this.fromChecked = fromChecked;
-            this.to = to;
-            this.toIncluded = toIncluded;
-            this.toChecked = toChecked;
-            this.reverse = reverse;
         }
 
         @Override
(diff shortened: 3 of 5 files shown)


classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -702,47 +706,39 @@ static class EntryIterator<K, V> implements TIterator<Entry<K, V>> {
...
-            TreeNode<K, V> down = node.down(reverse);
-            if (down != null) {
-                node = down;
-                while (node != null) {
-                    path[depth++] = node;
-                    node = node.forward(reverse);
-                }
-            }
+            last = path[path.length - 1];
+            path = owner.pathToNext(last.getKey(), reverse);
```
konsoletyper, line 721: It's a performance issue! You may not do so!

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -561,16 +563,16 @@ public Object clone() {
     }
 
     static class EntrySet<K, V> extends TAbstractSet<Entry<K, V>> implements TSequencedSet<Entry<K, V>> {
+        private final TTreeMap<K, V> owner;
```
konsoletyper, line 566: Changes like these give nothing, but make review harder

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
Ihromant: It could be, but there was failure in test_navigableKeySet with IndexOutOfBoundsException.

konsoletyper: So that does not mean this code is wrong. Perhaps, the code that maintains TreeNode.height is wrong

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -80,11 +80,11 @@ public void fix() {
         }
 
         public TreeNode<K, V> forward(boolean reverse) {
-            return !reverse ? left : right;
+            return reverse ? right : left;
```
konsoletyper, line 83: Not sure why you did this commit. I added inversion here intentially, since I believe that this code is *more* readable, when *normal* condition comes first.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
Ihromant: Restored back and fixed issue with removal.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
konsoletyper: Anyway, having bugfix and style commit together in one PR makes it difficult to review such PR

Ihromant: For me notinverted looks more readable. I'm assuming default value is false and looking to right when checking what value it should be. Also in other places it allowed to remove double inversions.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
Ihromant: Yes, it make review harder, but ordered firstly final and then not final fields improve readability significantly. Actually this readability helped alot to identify issues during fixing.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
Ihromant: I suppose that you can restore inverses after merging this PR in places you consider most important for readability.

konsoletyper: I don't agree with your point. The very meaning of this method is following. Consider we want to find next node in a binary tree. Then we move right and after that step move left until it's possible. Searching for the previous node is symmetric one: first move left and then move right until it's possible. Since these are symmetric, it was reasonable to generalize this algorithm. So what's a way to do this? Say that we always find "next" node, but there's option to search in forward or backward direction. For me, it's natural that we first formulate task as "forward" direction case, since it's easier to model in a head. So this "forward" case moves first. On the other hand, how should we call a flag that determines direction? For me it's natural to have "false" value for something by default, so there's no "forward" flag, but "reverse". This gives all these inversions. Perhaps, it's reasonable to have `forward(): boolean` method that simply inverts `reverse` flag.

There could be also concern that direction should not be encoded with boolean, rather with enum. However, reference to enum value takes 4 or 8 bytes, while `boolean` may occupy only one byte. Performance is quite important in library code.

konsoletyper: > I suppose that you can restore inverses after merging this PR in places you consider most important for readability.

No, why would I add extra garbage to git log? In case you refuse to revert that commit, I won't merge this PR, rather add my commits to this branch and them make squash

Ihromant: Reverted.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -753,36 +761,32 @@ public void remove() {
...
-            }
-            var newRoot = owner.deleteNode(owner.root, last.getKey());
-            if (owner.root != newRoot) {
-                owner.root = newRoot;
-                var newPath = owner.pathToNext(last.getKey(), reverse);
-                System.arraycopy(newPath, 0, path, 0, newPath.length);
-                depth = newPath.length;
+                throw new IllegalStateException();
             }
+            owner.root = owner.deleteNode(owner.root, last.getKey());
```
konsoletyper, line 766: This still a performance issue, not as bad as previous one. The path will be rebuilt after each deletion, however tree is not re-balanced after each deletion. The path should be only re-built after re-balancing, right? What actually caused AIOOBE, did you find out?

classlib/src/main/java/org/teavm/classlib/java/util/TAbstractMap.java
```diff
@@ -237,14 +237,18 @@ public boolean equals(Object obj) {
...
+            for (var it = entrySet().iterator(); it.hasNext();) {
+                TMap.Entry<K, V> entry = it.next();
+                if (!other.containsKey(entry.getKey())) {
+                    return false;
+                }
+                if (!TObjects.equals(entry.getValue(), other.get(entry.getKey()))) {
+                    return false;
+                }
             }
+        } catch (ClassCastException | NullPointerException e) {
```
konsoletyper, line 250: FYI: in JS BE these exceptions are thrown by VM *only* in strict mode, which is turned off by default, due to performance reasons. It's only turned on for tests. Usually it's a bad practice to rely on CCE and NPE as a normal contract, and *most* of the code I ever dealt with, follows this rule.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
Ihromant: Depth should be recalculated. AIOOBE was due to wrong depth after the deletion. For recalculating depth you should get new path. That was the reason why I removed "if" check. And I suppose that this is unavoidable.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
Ihromant: For me not. For instance, in MapView modCount and cachedSize were on the top, entrySetCache and cachedNavigableSet were on the bottom. It turned out that modCount and cachedize are not needed. Still, this is mess in my opinion. Also there were some wrong-order passing of variables, it was hard to guess about it without looking at properly ordered final fields. That is the reason for ordering them and distinguishing non-final from final.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
konsoletyper: Why should depth be recalculated? If three was not re-balanced, we have the same path leading to node being just removed. The next fetched node will always be searched from `depth - 1`, so it does not matter that `path[depth]` got changed. Even after re-balancing `path` length should be enough, since after deletion overall tree height can't grow.

classlib/src/main/java/org/teavm/classlib/java/util/TAbstractMap.java
Ihromant: I can explain to you this change. In contract for TreeMap and some other maps (ConcurrentHashMap for instance) it's possible to throw NPE or CCE during containsKey or get. equals method should never throw exceptions.
So, assuming the code
```
TreeMap<Integer, Integer> tm = new TreeMap();
tm.add(1, 1);
Map<String, String> m = Map.of("a", "b");
System.out.println(m.equals(tm));
```
in old code exception will be thrown instead of false. So either this fix or calls to other.containsKey() and other.get() should be guarded (via same try-catch) somehow from exceptions.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
konsoletyper: So I still *suspect* that there's an issue in the code that maintains tree height. If it's not the case and if you think that path recalculation is unavoidable, please, prove this. For example, you can describe sequence of steps that turn path into inconsistent state

Ihromant: Depth is not related to tree height.
Let's consider next tree:
```
       4 6  9
0 2   5   8
 1      7
     3
```
As you can see, `3` has path that consists of `[3]` only, `4` has path that consists of `[7, 5, 4]`. I understand that after deletion of 3 tree will be rebalanced, but you definitely can invent tree with two branches, where after deletion depth will grow.

classlib/src/main/java/org/teavm/classlib/java/util/TAbstractMap.java
Ihromant: Hmm, now get your point. In this case I assume that corresponding methods will return "false" and null. I'm assuming that in this case correct "false" will be returned to equals.

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
konsoletyper: > but you definitely can invent tree with two branches

Theorem, please!

My attempt for simple rotation:

Consider we have a subtree with root node X and balance factor +2. Let children of X are T1 and Y, children of Y are T2 and T3. height(Y) = height(T1) + 2, then height(T2) <= height(T1) + 1, height(T3) <= height(T1) + 1. height(X) = height(T1) + 3. 

After left turn we have Y1 as the root, where as X1 and T3 are its children, T1 and T2 are children of X. Then, height(X1)  = 1 + min(height(T1), height(T2)), but since height(T2) <= height(T1), min(height(T1), height(T2)) <= height(T1) + 2. height(Y1) = 1 + min(height(X), height(T3)), but min(height(X), height(T)) <= height(T1) + 2, so height(Y1) <= 3 + height(T1), Q.E.D.

Now you can write similar proof for double rotation.

Ihromant: There is even worse issue in code basically.
I restored failure in the code.
Added next method:
```java
private boolean fillSet(Set<K> set, TreeNode<K, ?> node) {
            if (set.add(node.getKey())) {
                System.out.println("Duplicate " + node.getKey());
                return false;
            }
            return fillSet(set, node.left) && fillSet(set, node.right);
        }
```
and wrapped IAOOBE into next protective catch:
```java
int lastDepth = depth;
            last = node;
            TreeNode<K, V> down = node.down(reverse);
            if (down != null) {
                node = down;
                while (node != null) {
                    try {
                        path[depth++] = node;
                    } catch (ArrayIndexOutOfBoundsException e) {
                        System.out.println(owner.root.height + " " + lastDepth + " " + last.height + " " + owner.size() + " " + (depth - 1) + " " + Arrays.toString(path));
                        Set<K> s = new HashSet<>();
                        System.out.println(fillSet(s, owner.root));
                        //BTreePrinter.printNode(owner.root);
                        throw e;
                    }
                    node = node.forward(reverse);
                }
            }
```
Output:
```java
11 8 6 983 11 [60=60, 35=35, 22=22, 19=19, 140=140, 13=13, 122=122, 121=121, 13=13, 126=126, 122=122]
Duplicate 60
false
```
It seems that TreeMap allows duplicates in some cases.

Ihromant: I tried to go other way. To print tree structure every iteration and find the place where bad structure happens. Unfortunately while System.out.println works in small cases, it for some reason doesn't work in iterations. It either ignores to go to output or goes to infinite loop where it shouldn't.

Ihromant: Steps to reproduce: 
https://github.com/konsoletyper/teavm/pull/813/commits/23ea0b7379162a23c85f7a8b6c701521d0c9eb6b#diff-959d6fa49c32ea239164890ed5b82a3f3fc97b3c787ad33585c6bd1be475673aR731
check this commit. Run, see that test fails (and see the output). Uncomment commented line `//System.out.println(depth + " " + path.length);`. Run test. Check for infinite loop. Comment this "try-catch" structure instead. Test now fails, but all System.out.println was ignored and nothing goes to output.

Ihromant: I tried to print it into console when launched test in IDEA during test `run` (not `debug`). No debugger, just System.out.println in this case.

konsoletyper: Is there any reason to not use debugger?

Ihromant: I wanted to have log in the end and look through it. Set size is 1000, failing case is 22th. I know about conditional breakpoints, but it is too complicated to prepare all for debugging every time you changing something.

konsoletyper: Quick hint for printing a tree: there's a popular tool for rendering graphs, called graphviz. It has its language for graph representation. There's GraphBuilder class in TeaVM whish creates Graph, and Graph.toString produces graphviz format.

Ihromant: Many thanks, will use it.
Still, this failure is not related to this copy-paste from StackOverflow, it's commented.

konsoletyper: In such cases solution is following: add verification method that throws exception when something gets inconsistent, then insert call to this method everywhere, then put exception breakpoint

konsoletyper: So I narrowed down problem. Looks like deletion can re-balance tree, and it's does not necessarily updates root. I'll investigate it further

konsoletyper: Ok, now I realized that it's impossible to incrementally update path (because it's not really whole path, but path containing only left-hand children), neither it's possible to even tell if path should be updated. But I can conclude, that idea of *not* keeping parent of each node was bad WRT performance. Even removal seems quite poor, requiring N*log(N) comparisons every time, even if it's performed from iterator. Perhaps, I'll refactor this later and simplify many things, i.e. size computation, navigation, iteration and so on.

Ihromant: By the way, what WRT acronym mean you are often using?

konsoletyper: with respect to = regarding, concerning

classlib/src/main/java/org/teavm/classlib/java/util/TAbstractMap.java
konsoletyper: Corresponding methods will produce UB (although there's not such thing in Java) like throwing native JS errors or returning values of unexpected types.

classlib/src/main/java/org/teavm/classlib/java/util/TNavigableMap.java
```diff
@@ -38,12 +38,16 @@ public interface TNavigableMap<K, V> extends TSortedMap<K, V> {
 
     K higherKey(K key);
 
+    @Override
```
konsoletyper, line 41: I suppose these methods are now inherited and can be removed?

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
konsoletyper: Please, revert these changes. It improves readability only from your standpoint. Another reason: I'd like to refactor TreeNode to contain link to parent and these fields will be changed. So these changes are nothing more than git history pollution

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -665,27 +678,22 @@ public boolean contains(Object o) {
...
 
     static class EntryIterator<K, V> implements TIterator<Entry<K, V>> {
-        private int modCount;
-        private TTreeMap<K, V> owner;
-        private TreeNode<K, V>[] path;
-        private TreeNode<K, V> last;
-        private K to;
-        private boolean toChecked;
-        private boolean toIncluded;
+        private final TTreeMap<K, V> owner;
```
konsoletyper, line 688: Please, revert this

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -753,36 +761,32 @@ public void remove() {
...
-        private int modCount = -1;
-        private int cachedSize;
-        private TTreeMap<K, V> owner;
-        private K from;
-        private boolean fromIncluded;
-        private boolean fromChecked;
-        private K to;
-        private boolean toIncluded;
-        private boolean toChecked;
+        private final TTreeMap<K, V> owner;
```
konsoletyper, line 776: Please, revert this

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -1265,6 +1303,18 @@ public TIterator<V> iterator() {
             };
         }
 
+        @Override
+        public boolean remove(Object o) {
+            for (TIterator<TMap.Entry<K, V>> it = map.entrySet().iterator(); it.hasNext();) {
+                TMap.Entry<K, V> e = it.next();
+                if (TObjects.equals(e.getValue(), o)) {
+                    map.remove(e.getKey());
```
konsoletyper, line 1311: Why not `it.remove()`?

konsoletyper:
I don't like the idea of committing this TreeMapTest1, but if you don't want to merge it with TreeMapTest, at least please resolve checkstyle issues.

Ihromant:
Could you please fix checkstyle in gradle?
When I'm running `./gradlew check` it fails with:
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not load the value of field `aliasSet` of `sun.nio.cs.UTF_8` bean found in field `charset` of `org.gradle.api.internal.resources.FileCollectionBackedTextResource` bean found in field `config` of task `:classlib:checkstyleMain` of type `org.gradle.api.plugins.quality.Checkstyle`.
> java.lang.reflect.InvocationTargetException (no error message)
```

konsoletyper:
What Java version are you using? Is it 21? Gradle does not support Java 21 yet, so you should run from 17

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -609,10 +618,14 @@ public int size() {
                         }
                     }
                     size -= path.length;
+                    if (fromPath != null && fromPath.length > 0 && path.length > 0 && owner.comparator.compare(
+                            fromPath[fromPath.length - 1].getKey(), path[path.length - 1].getKey()) == 0) {
```
konsoletyper, line 622: I suppose `fromPath[fromPath.length - 1] == path[path.length - 1]` is enough here

Ihromant:
Yes, it's 21.

Ihromant:
Still, it worked earlier on 21. But ok, will downgrade for tests.

konsoletyper:
I did not change anything significant here. The error is in Gradle, and it clearly states, that Java 21 is unsupported. Gradle doc also says that there's no support for running Gradle on Java 21 yet. I don't know what has changed, perhaps some random caching issues.

Ihromant:
Now when I'm compiling with 17 Java it fails with 
```
> Task :tests:compileTestJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':tests:compileTestJava'.
> error: invalid source release: 21
```

konsoletyper:
Please, read this article: https://docs.gradle.org/current/userguide/toolchains.html

Personally, I run Gradle from IDEA and specify JDK 17 as Gradle run JDK, but project has both 17 and 21 installed, so Gradle toolchains work properly. If it's not an option, I don't have a quick answer.

You can also try keep running on 21, but play with these lines in `gradle.properties`:

```
org.gradle.caching=true
org.gradle.unsafe.configuration-cache=true
```

especially, with the second one

classlib/src/main/java/org/teavm/classlib/java/util/TTreeMap.java
```diff
@@ -609,10 +618,14 @@ public int size() {
                         }
                     }
                     size -= path.length;
+                    if (fromPath != null && fromPath.length > 0 && path.length > 0
+                            && fromPath[fromPath.length - 1].getKey() == path[path.length - 1].getKey()) {
```
konsoletyper, line 622: getKey is also not necessary here

Ihromant:
Thanks, setting this properties helped. Now checkstyle and build are successful.

PR #818, opened 2023-10-12
Missing java.util static methods

Ihromant:
Missing java.util static methods


classlib/src/main/java/org/teavm/classlib/java/util/TCollections.java
```diff
@@ -460,6 +454,17 @@ public static int lastIndexOfSubList(TList<?> source, TList<?> target) {
         return -1;
     }
 
+    public static <T> TList<T> unmodifiableList(final TList<? extends T> list) {
```
konsoletyper, line 457: What was the reason to move this method? Such changes complicate review

Ihromant: Just grouped together all umodifiable... implementations.

PR #823, opened 2023-10-16
Small priority queue fixes

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/util/TPriorityQueue.java
```diff
@@ -112,6 +112,7 @@ public boolean offer(E e) {
         if (e == null) {
             throw new TNullPointerException();
         }
+        comparator.compare(e, e);
```
konsoletyper, line 115: What's the purpose of this call?

Ihromant: If we add not consistent element to queue of size 0 or 1 - it successfully adds and then we have broken state.

konsoletyper: What do you mean by "not consistent element"?

Ihromant: Ok, you're right, this check not needed.

PR #826, opened 2023-10-16
Speed up equals for Double and Float in JS

Ihromant:
Explanation:
```java
        Double[] array = Stream.generate(() -> Double.longBitsToDouble(ThreadLocalRandom.current().nextLong()))
                .limit(10_000_000).toArray(Double[]::new);
        int counter = 0;
        for (int i = 0; i < array.length - 1; i++) {
            counter += array[i].equals(array[i + 1]) ? 0 : 1;
        }
        for (int i = 0; i < array.length - 1; i++) {
            counter += array[i].equals(array[i + 1]) ? 0 : 1;
        }
        System.out.println("Time " + (System.currentTimeMillis() - time) + " counter " + counter);
    }
```
this simple benchmark shows:
Chrome:
Old: Time 406 counter 15001524
New: Time 94 counter 14991638
Firefox:
Old: Time 2390 counter 14991538
New: Time 142 counter 14990878
So, in JS we got 4 to 15 speed improvement by this simple change.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TDouble.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TDouble.java
@@ -16,6 +16,7 @@
 package org.teavm.classlib.java.lang;
 
 import org.teavm.backend.javascript.spi.InjectedBy;
+import org.teavm.classlib.PlatformDetector;
 import org.teavm.classlib.impl.text.DoubleSynthesizer;
 import org.teavm.interop.Import;
 import org.teavm.interop.NoSideEffects;
@@ -201,7 +202,9 @@ public class TDouble extends TNumber implements TComparable<TDouble> {
         if (this == other) {
             return true;
         }
-        return other instanceof TDouble && doubleToLongBits(((TDouble) other).value) == doubleToLongBits(value);
+        return other instanceof TDouble && PlatformDetector.isJavaScript()
+                ? TDouble.compare(((TDouble) other).value, value) == 0
+                : doubleToLongBits(((TDouble) other).value) == doubleToLongBits(value);
     }
 
     @Override
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TFloat.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TFloat.java
@@ -15,6 +15,7 @@
  */
 package org.teavm.classlib.java.lang;
 
+import org.teavm.classlib.PlatformDetector;
 import org.teavm.classlib.impl.text.FloatSynthesizer;
 import org.teavm.interop.Import;
 import org.teavm.interop.NoSideEffects;
@@ -85,7 +86,9 @@ public class TFloat extends TNumber implements TComparable<TFloat> {
         if (this == other) {
             return true;
         }
-        return other instanceof TFloat && floatToIntBits(((TFloat) other).value) == floatToIntBits(value);
+        return other instanceof TFloat && PlatformDetector.isJavaScript()
+                ? TFloat.compare(((TFloat) other).value, value) == 0
+                : floatToIntBits(((TFloat) other).value) == floatToIntBits(value);
     }
 
     @Override


Ihromant:
1. I didn't do benchmark for Float, but I assume that speedup will be lower, but still 2-7.
2. I don't think this will increase performance. Current toLongBits puts it in ArrayBufferView and then extract it. Proposed approach is similar to current implementation. Other issue here: NaN case, so code won't be simple.

konsoletyper:
> but I assume that speedup will be lower, but still 2-7.

Please, don't assume!

> I don't think this will increase performance. Current toLongBits puts it in ArrayBufferView and then extract it. 

Don't forget also that it deals with BigInt, which is slow by itself.

> Other issue here: NaN case, so code won't be simple.

What do you mean? `if (isNaN(value) && isNaN(that.value)) return true`?

Ihromant:
Updated benchmark for floats.
Chrome:
Old:
Doubles 444 counter 24997087
Floats 90 counter 34996938
New:
Doubles 93 counter 24995714
Floats 88 counter 34995576
Firefox:
Old:
Doubles 2458 counter 24995261
Floats 314 counter 34995094
New:
Doubles 144 counter 25000839
Floats 141 counter 35000696
Assumption was wrong. Still, in Chrome floats have approx. same speed while in Firefox change gives 2-times speedup, so improvement has benefit.

konsoletyper:
I suggest that you increase number of iterations by factor of at least 50, so the results will be relevant. Also I suggest to measure several times.

Ihromant:
Can't increase number of test cases. It starts crashing tabs even on 100_000_000. Still, added equal pairs.
```java
public class Client {
    private static final int TEST_COUNT = 10_000_000;
        Double[] array = Stream.generate(() -> Double.longBitsToDouble(ThreadLocalRandom.current().nextLong()))
                .limit(TEST_COUNT).toArray(Double[]::new);
        Float[] fArray = Stream.generate(() -> Float.intBitsToFloat(ThreadLocalRandom.current().nextInt()))
                .limit(TEST_COUNT).toArray(Float[]::new);
        IntStream.generate(() -> ThreadLocalRandom.current().nextInt(array.length)).limit(array.length / 4).forEach(i -> {
            array[i] = 5.0;
            fArray[i] = 3.0f;
        });
        int counter = 0;
        for (int i = 0; i < array.length - 1; i++) {
            counter += array[i].equals(array[i + 1]) ? 0 : 1;
            counter += fArray[i].equals(fArray[i + 1]) ? 0 : 1;
        }
        for (int i = 0; i < array.length - 1; i++) {
            counter += array[i].equals(array[i + 1]) ? 0 : 1;
        }
        System.out.println("Doubles " + (System.currentTimeMillis() - time) + " counter " + counter);
        time = System.currentTimeMillis();
        for (int i = 0; i < fArray.length - 1; i++) {
            counter += fArray[i].equals(fArray[i + 1]) ? 0 : 1;
        }
        System.out.println("Floats " + (System.currentTimeMillis() - time) + " counter " + counter);
    }
}
```
Results:
Chrome:
Old:
Doubles 497 counter 25493442
Floats 154 counter 35003106
Doubles 518 counter 25494665
Floats 154 counter 35005594
Doubles 505 counter 25496297
Floats 153 counter 35007874
New:
Doubles 167 counter 25496649
Floats 154 counter 35007636
Doubles 167 counter 25499776
Floats 154 counter 35010534
Doubles 170 counter 25501053
Floats 156 counter 35012212
Firefox:
Old:
Doubles 4798 counter 25497246
Floats 618 counter 35007078
Doubles 4829 counter 25500967
Floats 578 counter 35011228
Doubles 4775 counter 25501696
Floats 582 counter 35013302
New:
Doubles 393 counter 25499172
Floats 372 counter 35010210
Doubles 408 counter 25498895
Floats 404 counter 35010110
Doubles 392 counter 25499503
Floats 380 counter 35010690

Same conclusion, same speed for floats in Chrome, but improvement for other cases.

konsoletyper:
You don't need to waste memory. You can wrap working loops and repeat then, say, 10 times

classlib/src/main/java/org/teavm/classlib/java/lang/TDouble.java
```diff
@@ -201,7 +202,9 @@ public boolean equals(Object other) {
         if (this == other) {
             return true;
         }
-        return other instanceof TDouble && doubleToLongBits(((TDouble) other).value) == doubleToLongBits(value);
+        return other instanceof TDouble && PlatformDetector.isJavaScript()
```
konsoletyper, line 205: This is not correct condition. You must check if other is TDouble, and then perform platform type check. What actually happens: if you have other is Double *and* platform is JavaScript, use JavaScript-specific comparison, otherwise perform IEEE754 comparison. So actually passing something else than TDouble would result in CCE.

classlib/src/main/java/org/teavm/classlib/java/lang/TDouble.java
```diff
@@ -201,7 +202,9 @@ public boolean equals(Object other) {
...
+                ? TDouble.compare(((TDouble) other).value, value) == 0
```
konsoletyper, line 206: Instead of comparison it's also possible to do following:

```java
a == b && (1 / a) == (1 / b) || a != a && b != b
```

which I believe a bit faster

Issue #735, opened 2023-08-28
Double.parseDouble(Double.toString( number )) != number 


McMurat:
Because of the current implementation of Double.parseDouble and Double.toString, it is not possible to serialize data without change of data representation. I know that implementing 100% correct parseDouble/toString methods is challenging, but maybe it would be a good idea to only check the format and then delegate the work to JavaScript. Although this would not work for WASM/C.

konsoletyper:
Also JavaScript format requires marshalling between JS and Java, which is expensive. Could you please provide examples of such cases?

McMurat:
One number returned by Double.toString() is 74.92507492507494
When I parse this using Double.parseDouble() and  convert it toString() again then I get 74.92507492507505

The difference is pretty small but problematic in my case as I serialize my internal data structures to Strings and compare them to check for differences and there are differences although there should not be any.

McMurat:
Another number which I also inspected in Chrome DevTools:

39.7291655739246 (Chrome Dev Tools number field)
39.7291655739246 (String representation, TeaVM toString())
39.729165573924654 (TeaVM parseDouble and then toString() again)

Seems like the data loss is in parseDouble as toString seems to give the same representation (at least in the cases I checked).

konsoletyper:
@McMurat I can confirm that the problem in the parser. I tried another algorithm with direct IEEE 754 construction instead of floating point operations (like toString currently works) and this gives me better result, it also fixes `74.92507492507494` case. I'll push my changes as soon I test this more and implement same approach for Float.parseFloat.

McMurat:
Excellent work, I will try it out as soon as it is pushed.

konsoletyper:
@McMurat e2ee9f1dbba7848b1b5cc7475a26ef448d6acfc5 addresses this issue, however both parser and formatter aren't perfect now. I tried approach with comparison of random numbers, and after this commit only 0.5% doubles and 3.5% floats have inconsistent parse/format results (the difference is always by lower bit in binary mantissa). This is much better than before commit. I don't know why there are inconsistencies, I'm not that strong at math. However, I won't use JS parse/format functions for number of reasons:

1. I want TeaVM to provide consistent result, not depending on browser version.
2. I want TeaVM to provide consistent result between targets (wasm, C, JS).
3. There's also formatted parser and formatter, which share same algorithm.

[...]

PR #829, opened 2023-10-22
Initial re-implementation for Double.parseDouble/Float.parseFloat

Ihromant:
Hi @konsoletyper . This is initial (and therefore unfinished) implementation for Double.parseDouble improvement. Code was borrowed from https://android.googlesource.com/platform/libcore/+/d5e2817/luni/src/main/java/java/lang/Double.java Android source repository (Apache 2.0). Test was borrowed from there as well. So, test shows that we have issues in Double.toHexString (suppose will be fixed separately) and with parsing. I reused implementation and replaced native implementation to the most naїve approach instead of real implementation (should be transformed from C++ https://cs.android.com/android/platform/superproject/+/master:libcore/luni/src/main/native/java_lang_StringToReal.cpp;l=92;drc=master ). Is it worth continuing working on this PR or stop here?
It appears that even naїve implementation covers most of the tests that currently present.


konsoletyper:
1. There's already existing implementation that works quite well except for few corner cases. Isn't it better to fix it rather than to replace everything by another implementation?
2. Current implementation of Double.parseDouble uses common engine to turn decimal exponent + decimal mantissa into IEEE 754 binary exponent + binary mantissa (see DoubleSynthesizer). The same engine used by TDecimalFormat, so whenever you change something, you should also take care of TDecimalFormat
3. This algorithm uses naive approach that was implemented in TeaVM quite some time ago. Looks like acting in terms of IEEE 754 doubles does not provide enough accuracy, so I replaced everything with integer math. Using long to store mantissa gives 64 bits instead of 52 bits. 
4. This is not only inaccurate, but also slow approach. Computing 10^x every time may be slow compared to using tables.
5. Recently, I improved my implementation to use compile-time-calculated 100%-precise tables to convert between decimal and binary mantissa (I used BigInteger, then cut first 64 bits). This gave much far better result, not 100% accurate though.
6. You should test any changes there WRT consistency between parseDouble/Double.toString (i.e. feed result of Double.toString to parseDouble for large amount of random numbers). After my recent changes there's approx. 0.2% of numbers that give only 1 ULP inaccuracy (i.e. binary mantissa differs only by 1). Did you test your changes this way?

Ihromant:
1. Here is your comment https://github.com/konsoletyper/teavm/issues/735#issuecomment-1718063424 . According to it you managed to produce 99,5% and 96,5% accurate parse for doubles/floats respectively. This just means that it's (I'm not saying impossible) complicated and unneeded to spend efforts to improve custom self-written algorithm. It's better to reuse common knowledge shared under Open-Source-friendly license. Together with the code it comes with applicable tests. I understand that it's pretty challenging, but it's not worth your/my time.
2. Thanks for the info, I didn't consider it in beginning. Still, formatting is about toString, not parsing. Then I suppose that changes for toString also should be included to be consistent. In Android test suite only 4 tests failed in JDK, so I'm assuming that this implementation is pretty similar and precise.
3. This naїve approach was used just to make code working. Check TODO link. There is C++ code that uses same tables approach under the hood. It just needs to be rewritten in Java.
4. See 3.
5. See 3.
6. Sure, it should be tested well, basic question is unanswered: whether continue working on this PR or just drop it? I didn't test my changes WRT consistency because I don't want to prepare full PR (this is not 1-day work obviously) which later will be refused.

konsoletyper:
> This just means that it's (I'm not saying impossible) complicated and unneeded to spend efforts to improve custom self-written algorithm.

No, I can't agree. At least, after all my efforts I can't let you throw away my almost perfectly working code.

> It's better to reuse common knowledge shared under Open-Source-friendly license.

Yes, and get the code that neither you, nor I understand. And moreover, you need to somehow port it from C++.

> Sure, it should be tested well, basic question is unanswered: whether continue working on this PR or just drop it

In this form drop it. You can take some tests from Harmony/Android implementation to find issues in current implementation, as well as fix these errors in current code. You can read some papers about decimal-to-binary conversion to improve accuracy and fix current algorithm. But in this current it's almost impossible to review.

PR #831, opened 2023-10-22
Character fixes

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/lang/TCharacter.java
```diff
@@ -177,7 +178,7 @@ public static boolean isSurrogatePair(char high, char low) {
     }
 
     public static int charCount(int codePoint) {
-        return isSupplementaryCodePoint(codePoint) ? 2 : 1;
+        return codePoint >= MIN_SUPPLEMENTARY_CODE_POINT ? 2 : 1;
```
konsoletyper, line 181: Why?

classlib/src/main/java/org/teavm/classlib/java/lang/TCharacter.java
```diff
@@ -199,6 +200,9 @@ public static int codePointAt(char[] a, int index) {
     }
 
     public static int codePointAt(char[] a, int index, int limit) {
+        if (index >= limit || index < 0 || limit > a.length) {
```
konsoletyper, line 203: Objects.checkFromToIndex?

konsoletyper: Ah, a little bit other comparison

classlib/src/main/java/org/teavm/classlib/java/lang/TCharacter.java
Ihromant: Because charCount(Integer.MAX_VALUE) for instance is 2 (despite it's meaningless).

PR #834, opened 2023-10-24
Enum set map fixes

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
```diff
@@ -16,44 +16,34 @@
...
 import java.util.Arrays;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Objects;
-import java.util.Set;
+import org.teavm.classlib.java.lang.TCloneNotSupportedException;
+import org.teavm.interop.Rename;
 
-public class TEnumMap<K extends Enum<K>, V> extends AbstractMap<K, V> implements Serializable, Cloneable {
```
konsoletyper, line 28: Personally, I have another trend: to avoid using these `T`-prefixed classes when possible.

konsoletyper: Anyway, these changes make reviews harder

classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
```diff
@@ -16,44 +16,34 @@
...
     private Class<K> keyType;
     private Object[] data;
     private boolean[] provided;
     private int size;
-    private Set<Entry<K, V>> entrySet;
+    private TSet<Entry<K, V>> entrySet;
 
     public TEnumMap(Class<K> keyType) {
-        initFromKeyType(keyType);
+        initFromKeyType(TObjects.requireNonNull(keyType));
```
konsoletyper, line 31: I think initFromKeyType will throw NPE if keyType is null

classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
```diff
@@ -16,44 +16,34 @@
...
                 throw new IllegalArgumentException();
             }
             initFromKeyType(m.keySet().iterator().next().getDeclaringClass());
-            for (Entry<K, V> entry : m.entrySet()) {
-                int index = entry.getKey().ordinal();
-                provided[index] = true;
-                data[index] = entry.getValue();
-            }
-            size = m.size();
+            putAll(m);
```
konsoletyper, line 46: Isn't old implementation more efficient?

classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
```diff
@@ -107,6 +97,10 @@ public V get(Object key) {
 
     @Override
     public V put(K key, V value) {
+        Class<?> cls = key.getClass();
+        if (cls != keyType && cls.getSuperclass() != keyType) {
```
konsoletyper, line 101: I agree that this is right, but things like these don't add much business value to TeaVM. What is your motivation to propose such changes?

classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
```diff
@@ -209,13 +229,31 @@ public int size() {
                     return size;
                 }
 
+                @Override
+                public boolean contains(Object o) {
+                    if (!(o instanceof TMap.Entry<?, ?>)) {
+                        return false;
+                    }
+                    TMap.Entry<?, ?> e = (TMap.Entry<?, ?>) o;
+                    if (!keyType.isInstance(e.getKey())) {
```
konsoletyper, line 238: In the code above you used following check:

```
cls != keyType && cls.getSuperclass() != keyType
```

classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
Ihromant: My motivation was to fix all found issues in Harmony tests.

classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
Ihromant: It fails when we put some map like Map.of(Size.small, 1, Color.green, 2)

classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
Ihromant: This change was forced by considering specific TEnumMap case for putAll.

classlib/src/main/java/org/teavm/classlib/java/util/TEnumMap.java
konsoletyper: So may be it was reasonable to fix old implementation?

tests/src/test/java/org/teavm/classlib/java/util/EnumSetTest.java
```diff
@@ -196,6 +196,214 @@ public void retainAll() {
         assertEquals(original, set);
     }
 
+    @Test
+    public void iterator() {
```
konsoletyper, line 200: This test does not pass in WebAssembly in optimized mode. This is likely to some bug, not related to this particular commit, so please just suppress WebAssembly and WASI targets for this test

PR #838, opened 2023-10-27
Long fixes

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/lang/TLong.java
```diff
@@ -358,10 +375,8 @@ public static long reverse(long i) {
     }
 
     public static long reverseBytes(long i) {
-        i = ((i & 0xFF00FF00FF00FF00L) >> 8)  | ((i & 0x00FF00FF00FF00FFL) << 8);
-        i = ((i & 0xFFFF0000FFFF0000L) >> 16) | ((i & 0x0000FFFF0000FFFFL) << 16);
-        i = (i >> 32) | (i << 32);
-        return i;
+        i = (i & 0x00ff00ff00ff00ffL) << 8 | (i >>> 8) & 0x00ff00ff00ff00ffL;
```
konsoletyper, line 378: Why didn't you just replace `>>` with `>>>` in existing implementation?

classlib/src/main/java/org/teavm/classlib/java/lang/TLong.java
```diff
@@ -71,90 +104,74 @@ private static long parseLongImpl(CharSequence s, int beginIndex, int endIndex,
...
-        }
-        int radix = 10;
-        if (nm.charAt(index) == '#') {
-            radix = 16;
-            ++index;
-        } else if (nm.charAt(index) == '0') {
-            ++index;
-            if (index == nm.length()) {
-                return TLong.valueOf(0);
+    public static TLong decode(String s) throws TNumberFormatException {
```
konsoletyper, line 113: Is it just a copy from Harmony? Why didn't you just fixes issues in existing implementation?

Ihromant: It's not a dumb copy, but adapration due to:
1. Harmony doesn't consider cases like "+#abc".
2. ParseLong(Charsequence, int, int, int) is missing in Harmony at all.

Reason:
for me it was easier to adapt Harmony implementation than dig into code and search for issue there.

konsoletyper: > for me it was easier to adapt Harmony implementation than dig into code and search for issue there.

For me it's easier to review code that has actual fixes instead of rewriting everything

konsoletyper: For me it is

Ihromant: Current code repeats same parsing (actually by like copy-pase), therefore it contains same bugs in parseInt and decode. New implementation moves common part (parsing non-negative meaningful part) to parseImpl after some preliminary work. I don't think that it's possible to make this fix simple without this "rewrite".

konsoletyper: Quote from Javadocs:

> DecimalNumeral, HexDigits, and OctalDigits are as defined in section [3.10.1](https://docs.oracle.com/javase/specs/jls/se21/html/jls-3.html#jls-3.10.1) of The Java Language Specification, except that underscores are not accepted between digits.

> The sequence of characters following an optional sign and/or radix specifier ("0x", "0X", "#", or leading zero) is parsed as by the Long.parseLong method with the indicated radix (10, 16, or 8)

According to Java spec, only latin numbers accepted, however the second paragraph states that number is actually parsed by Long.parseLong, which accepts any digits. So it looks peculiar to me.

konsoletyper: Anyway, there was no reason to re-write part of the method that goes before calling "parseImpl"

Ihromant: I suppose that we can accept only latin numbers by restoring private decodeNumber. It will be better from performance standpoint and nobody needs parsing some indian/arabic digits.

Ihromant: > Anyway, there was no reason to re-write part of the method that goes before calling "parseImpl"

I can explain to you logic of this refactoring step-by-step.
1. Exceptions/validations in current implementation are incorrect. They should be separate for all 3 methods that eventually call parseImpl.
2. ParseImpl does the same thing as before. Difference in implementation is  the following: instead of going through positive numbers, it goes through nevative numbers (because there is more negative). In the end when we obtain MIN_VALUE - we return it, otherwise it fails. Yes, code changes are large, but in the nutshell, this changes are very simple and I understand every steps of this code and why it handles Integers (in neighboring PR) and Longs correctly.

konsoletyper: > Exceptions/validations in current implementation are incorrect. They should be separate for all 3 methods that eventually call parseImpl.

There are lots of things beyond these exceptions/validations. Sometimes logic is identical, but variables are renamed.

> Difference in implementation is the following: instead of going through positive numbers, it goes through nevative numbers (because there is more negative)

I can't agree with this. This ancient code was written in pre-8 era, when there were no unsigned comparisons.

Ihromant: Still, it's effective enough and covers all cases. Current implementation has issues with numbers with same number of digits as MIN_VALUE.

konsoletyper: It's ridiculous. Why using some ancient hack when there's Long.compareUnsigned?

Ihromant: Please comment the exact place. For me it's unclear where Long.compareUnsigned should be put.

PR #843, opened 2023-10-28
Small string fixes

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
```diff
@@ -418,20 +428,20 @@ public int lastIndexOf(TString str) {
...
-        if (PlatformDetector.isJavaScript()) {
-            var nativeSubstring = substringJS(nativeString(), beginIndex, endIndex);
-            return nativeSubstring != nativeString() ? new TString(nativeSubstring) : this;
+        if (beginIndex == 0 && endIndex == length) {
+            return this;
         }
 
-        if (beginIndex == 0 && endIndex == length()) {
-            return this;
+        if (PlatformDetector.isJavaScript()) {
```
konsoletyper, line 443: With this I can't agree. Previous code seems better

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
```diff
@@ -122,6 +127,7 @@ public TString(int[] codePoints, int offset, int count) {
     }
 
     private void initWithBytes(byte[] bytes, int offset, int length, TCharset charset) {
+        checkSIOOB(offset, length, bytes.length);
```
konsoletyper, line 130: Is this check really necessary? Won't further code throw some exception when indexes are invalid?

classlib/src/main/java/org/teavm/classlib/java/util/TObjects.java
```diff
@@ -158,14 +158,14 @@ public static long checkFromToIndex(long fromIndex, long toIndex, long length) {
     }
 
     public static int checkFromIndexSize(int fromIndex, int size, int length) {
-        if (fromIndex < 0 || size < 0 || fromIndex + size > length) {
+        if (fromIndex < 0 || size < 0 || size > length - fromIndex) {
```
konsoletyper, line 161: What was the reason for this change?

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
```diff
@@ -80,30 +81,34 @@ static TString fromArray(char[] characters) {
     private native void takeCharArray(char[] characters);
 
     public TString(byte[] bytes, int offset, int length, TString charsetName) throws TUnsupportedEncodingException {
-        this(bytes, offset, length, TCharset.forName(charsetName.toString()));
+        checkSIOOB(offset, length, bytes.length);
```
konsoletyper, line 84: initWithBytes does exactly the same in its first line

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
Ihromant: Why? Strings are equal only in case when beginIndex == 0 and endIndex == length. Removed comparison makes no sense when this values are different.

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
konsoletyper, line 84: I want also point out that there's hardly any code that depend on such checks. But checks like these have impact on performance and generated code size

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
Ihromant: It will throw (and throws), but exception is incorrect.

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
konsoletyper: So let JavaScript do this check

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
konsoletyper: In 99.99999% of situations there's no difference at all. I can hardly imagine some code, except for synthetic tests for JVM spec conformance, that really relies on SIOOB/AIOOB exception type and makes such difference.

konsoletyper: As an alternative you can also wrap AIIOB (or whatever exception is thrown) and re-throw SIOOB. Exceptions, that aren't thrown, have zero cost. There's still some impact on JavaScript size though.

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
Ihromant: You can't never be sure. In my practice I have seen lots of code in different libraries which instead of doing preliminary checks, just does something like
```java
try {
// do stuff
} catch (IndexOutOfBoundsException e) {
// throw own wrapped exception
}
```
I understand that by default in teavm this checks are relaxed, but somebody can compile in strict mode when he relies on this libraries.

konsoletyper: It's still IOOBE.

classlib/src/main/java/org/teavm/classlib/java/util/TObjects.java
Ihromant: Corner cases like length = Integer.MAX_VALUE, size = Integer.MAX_VALUE, fromIndex = 2.

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
Ihromant: I'm assuming that comparing two integers is faster in most cases than comparing 2 strings.

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
Ihromant: No, JDK throws StringIndexOutOfBoundsException. I'm not insisting on this change, but see comment above (with `catch StringIndexOutOfBoundsException`.

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
Ihromant: So you are suggesting to use the same `Objects.checkFromIndexSize`?

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
konsoletyper: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html#%3Cinit%3E(char%5B%5D,int,int)

No mentions of this SIOOBE except for indexOf

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
konsoletyper: No, I suggest not to perform any check at all. The line below uses ByteBuffer.wrap, which performs the very same check: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/ByteBuffer.html#wrap(byte%5B%5D,int,int)

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
```diff
@@ -80,30 +81,34 @@ static TString fromArray(char[] characters) {
...
+        initWithBytes(bytes, 0, bytes.length, TCharset.forName(charsetName.toString()));
     }
 
     public TString(byte[] bytes, TCharset charset) {
-        this(bytes, 0, bytes.length, charset);
+        initWithBytes(bytes, 0, bytes.length, charset);
     }
 
     public TString(int[] codePoints, int offset, int count) {
+        Objects.checkFromIndexSize(offset, count, codePoints.length);
```
konsoletyper, line 111: The code below reads from the array, which in strict mode will throw AIIOBE. So is this check really necessary?

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
Ihromant: Added missing check to ByteBuffer instead and removed everywhere where bytes are used.

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
```diff
@@ -418,20 +423,20 @@ public int lastIndexOf(TString str) {
     }
 
     public TString substring(int beginIndex, int endIndex) {
-        if (beginIndex > endIndex) {
-            throw new TIndexOutOfBoundsException();
+        int length = charactersLength();
+        if (beginIndex < 0 || beginIndex > endIndex || endIndex > length) {
```
konsoletyper, line 427: I think this check is only necessary for JS. In other BEs TString constructor will perform the same check

classlib/src/main/java/org/teavm/classlib/java/lang/TString.java
konsoletyper: I assume that native comparison with either have two referentially identical string and returns `true` by the fast path, or otherwise deals with strings of different sizes, so also avoids full char-by-char check. You can also write a benchmark that tests if there's any difference before and after your change.

Ihromant: In this case still 2 "integer" comparisons (first references and second length) are performed.
But ok, I will write benchmark.

Ihromant:
```java
private static final int TEST_COUNT = 1000_000;
        String[] array = Stream.generate(() -> UUID.randomUUID().toString()).limit(TEST_COUNT).toArray(String[]::new);
        BitSet bs = BitSet.valueOf(LongStream.range(0, TEST_COUNT / Long.BYTES + 1).map(i -> ThreadLocalRandom.current().nextLong()).toArray());
        int counter = 0;
        for (int i = 0; i < array.length - 1; i++) {
            counter += bs.get(i) ? array[i].length() : array[i].length() - 1;
        }
        for (int j = 0; j < 1000; j++) {
            for (int i = 0; i < array.length - 1; i++) {
                counter += bs.get(i) ? array[i].substring(0, 36) == array[i] ? 1 : 0
                        : array[i].substring(0, 35) == array[i] ? 1 : 0;
            }
        }
        System.out.println("Comparisons " + (System.currentTimeMillis() - time) + " counter " + counter);
    }
```
Luckily for me new implementation is little bit, but faster. Seems that 2 times comparing integers works faster than reference arithmetics/dereferencing. Results:
|`Browser `|`old substring  `|`new substring  `|
|`Chromium`|`14335,14313,14202`|`12098,11971,11802`|
|`Firefox `|`29022,29476,29179`|`26288,27397,26150`|

PR #844, opened 2023-10-28
Update DefaultGlobalNameWriter.java to also include dynamic `import` …

aghasemi:
…in non-appendable functions


konsoletyper:
This PR is simply wrong. It solves one particular case, but in general it's wrong. Instead, TeaVM needs to support ES6 modules (which includes ES6 syntax in general), needs to get rid of this `$rt_globals` (which actually a hack), but a big refactoring is required, which I'm currently working on.

aghasemi:
> This PR is simply wrong. It solves one particular case, but in general it's wrong. Instead, TeaVM needs to support ES6 modules (which includes ES6 syntax in general), needs to get rid of this `$rt_globals` (which actually a hack), but a big refactoring is required, which I'm currently working on.

I expected this to be incorrect. Just wanted to open the discussion. Looking forward to the refactoring.

PR #853, opened 2023-11-07
Widen byte/short to ints in ObjectMethods

SquidDev:
If you have a record a `byte` (or `short`) field, TeaVM would generate a `toString` method which called the (non-existent) `StringBuilder.append(B)`. We need to do similar logic to [what `StringBuilderEmitter` does](https://github.com/konsoletyper/teavm/blob/8f8b33c0b1a78292dc4a438ff42ac42e75e59663/core/src/main/java/org/teavm/model/emit/StringBuilderEmitter.java#L32-L40) and convert byte/shorts to `int`s.

### DIFF
--- a/tests/src/test/java/org/teavm/vm/RecordTest.java
+++ b/tests/src/test/java/org/teavm/vm/RecordTest.java
@@ -42,7 +42,7 @@ public class RecordTest {
 
     @Test
     public void toStringMethod() {
-        String s = new B(2, "q", 3L).toString();
+        String s = new B(2, "q", 3L, (byte) 4, (short) 5).toString();
 
         int index = 0;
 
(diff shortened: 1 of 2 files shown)


tests/src/test/java/org/teavm/vm/RecordTest.java
```diff
@@ -74,9 +74,18 @@ public void toStringMethod() {
         assertTrue(index > 0);
     }
 
+    @Test
+    public void toStringMethodPrimitives() {
+        String s = new AllPrimitives(false, (byte) 1, '2', (short) 3, 4, 5L, 6.0f, 7.0, "test").toString();
+        assertEquals("AllPrimitives[a=false, b=1, c=2, s=3, i=4, l=5, f=6.0, d=7.0, o=test]", s);
```
konsoletyper, line 80: I usually don't write record tests like this. There's no specific format in spec, so it may change even across JVM versions, making such tests unstable.

SquidDev: Ahh, that's a good point!

I'm not sure it's worth having another version of the existing `toString` test, so could either add all the primitives to the `B` record (and test), or just remove this `toStringMethodPrimitives` test. Do you have any preference?

konsoletyper: Yes, adding byte and short values to B is a good idea

PR #862, opened 2023-11-15
Fix `toLowerCase`/`toUpperCase` on `Locale.ROOT`

SquidDev:
`Locale.ROOT.toLanguageTag()` currently evaluates to an empty string, rather than the expected "und", which errors when passed to `toLocaleLowerCase`.


konsoletyper:
I believe, this case should also be covered in `LocaleTest`

PR #864, opened 2023-11-23
Extracted not added items from PR #313

Ihromant:
https://github.com/konsoletyper/teavm/pull/313/files
lots of items already were implemented, so I extracted everything that was not implemented. Did not merge it, but I suppose that is should be authored by @niloc132

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/dom/types/DOMTokenList.java
+++ b/jso/apis/src/main/java/org/teavm/jso/dom/types/DOMTokenList.java
@@ -30,7 +30,17 @@ public interface DOMTokenList extends JSObject {
 
     void add(String token);
 
+    void add(String token1, String token2);
+
+    void add(String... tokens);
+
     void remove(String token);
 
+    void remove(String token1, String token2);
+
+    void remove(String... tokens);
+
     boolean toggle(String token);
+
+    boolean toggle(String token, boolean force);
 }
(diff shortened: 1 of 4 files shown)


jso/apis/src/main/java/org/teavm/jso/dom/types/DOMTokenList.java
```diff
@@ -30,7 +30,17 @@ public interface DOMTokenList extends JSObject {
 
     void add(String token);
 
+    void add(String token1, String token2);
+
+    void add(String... tokens);
```
konsoletyper, line 35: Are you sure that varargs supported in JSO?

Ihromant: There also was a comment to create issue to support varargs in JS. Ok, will create issue to support varagrs.

konsoletyper: So are they supported or not? If they aren't, I can't accept PR with these declarations

Ihromant: I thought that you know as author of TeaVM. But ok, will check.

konsoletyper: I'm not a machine remebering each and every character from this repository

Ihromant: Sorry, didn't want to offend you. Just checked. Varargs are not supported. So, extracted varargs to separate PR https://github.com/konsoletyper/teavm/pull/867

PR #868, opened 2023-11-24
Various fixes for Arrays

Ihromant:


Ihromant:
Failures are related to ByteBuffer in wasi, it changed nothing there.

konsoletyper:
Did you try to debug failures in WASI?

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
```diff
@@ -37,84 +39,90 @@
 
 public class TArrays extends TObject {
     public static char[] copyOf(char[] array, int length) {
-        char[] result = new char[length];
         int sz = TMath.min(length, array.length);
-        for (int i = 0; i < sz; ++i) {
-            result[i] = array[i];
+        if (length < 0) {
```
konsoletyper, line 43: This shoud not be done here. Line below instantiates array and JVM should throw exception there.

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
```diff
@@ -123,92 +131,111 @@ public static <T, U> T[] copyOf(U[] original, int newLength, Class<? extends T[]
         Class<?> componentType = cls.getComponentType();
         T[] result = (T[]) Array.newInstance(componentType, newLength);
         int sz = TMath.min(newLength, original.length);
-        for (int i = 0; i < sz; ++i) {
-            result[i] = (T) componentType.cast(original[i]);
-        }
+        TSystem.arraycopy((TObject) (Object) original, 0, (TObject) (Object) result, 0, sz);
```
konsoletyper, line 134: System?

Ihromant: This is made intentionally. In exactly this case no fastArrayCopy optimization should be applied.

konsoletyper: Why?

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
Ihromant: JVM don't throw required exception. I investigated create array instructions and they indeed just create array without any check (and we get error in JS). Still, I don't know how to add preliminary check and throw. Maybe some example?

Ihromant:
I don't know how to launch and debug WASI, in my config this tests are just off.

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
Ihromant: Take a look at https://github.com/konsoletyper/teavm/pull/868/files#diff-58413b8b823bf4fc7635dfcc153675834256930684714088015dffe99f63ee34R299

konsoletyper: Ok, if TeaVM optimizes out some check it's not supposed to optimize out, then may be you investigate issue in the optimizer?

konsoletyper:
What's the problem to **learn** to debug WASI? This "I don't know" is never a good excuse

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
konsoletyper: https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5.anewarray

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
Ihromant: There is no difference between copyOf(T[] arr, int from, int to) and copyOf(T[] arr, int from, int to, Class<U> cl) from bytecode standpoint. They both are Object[] in bytecode, so optimizer will make no difference. Still, in first case we know that fastArrayCopy is safe while in second case we don't.

classlib/src/main/java/org/teavm/classlib/java/util/TArrays.java
Ihromant: https://github.com/konsoletyper/teavm/blob/master/core/src/main/java/org/teavm/parsing/ProgramParser.java#L478 here this anewarray is converted to instruction. But I don't see in code where length checks and throws are added then.

Ihromant:
From my earlier experience with C/WASM backend, it's either "you try some magic and pray for it working" (what I previously did) or spend 2-3 months of learning this platform to understand what's happening (which I don't have unfortunately).

Ihromant:
@konsoletyper seems that after your recent changes in buffers, two platforms started failing instead of one in this PR (why???). It could point to some deeply hidden bug on WASM/WASI platforms that I don't understand. Changes here are straightforward (and optimizing), but they continue failing exactly on WASI platrofm for very long time.

konsoletyper:
@Ihromant you tell me

Ihromant:
There is no copyOf/copyOfRange direct usages in T*primitive*Buffer/TBuffer at all. It could be somewhere in intrinsification, but I can't find any place where it is applied. The only suspicious place is Address.moveMemoryBlock() invocation (just because it has similar signature), but I don't see how it's related to Arrays. So for me it continues to be mystery that prevents from finishing this PR.

konsoletyper:
@Ihromant I have no idea what you talking about. Which `copyOf` and how are they related to the issue?

Ihromant:
Changes in this PR are in `Arrays.copyOf` / `Arrays.copyOfRange`. But tests that are failing in WASI/WASM platforms according to the logs are T*primitive*BufferTest. Tests for `copyOf` / `copyOfRange` are successfully passed in the same time.

Ihromant:
Also on WASM ByteBufferTest is passed successfully and on WASI it fails. This adds even more mystery to this issue.

konsoletyper:
@Ihromant didn't it occur to you that I'm not going to merge this PR?

Ihromant:
@konsoletyper I understand that, but this PR contains valid fixes for which tests were added (and which are failing in current `master` branch).

konsoletyper:
@Ihromant again this "valid". There are parts of JVM spec that are too costly too implement, and TeaVM tries to be pragmatic, i.e. intentionally omits some corner cases in favour of performance. So in this particular context these fixes and these tests are not valid.

PR #871, opened 2023-12-02
Initial commit for negative array size check

Ihromant:
Hi @konsoletyper 

Here is initial implementation for negative array size check. I investigated code and assumed the place where instructions should be modified. I can see that in BoundCheckInsertion class has code that intercepts ConstructArrayInstruction in checks and then if check is correct then restores flow of the program, else throws corresponding exceptions. Still, I faced problems to solve which I added TODO comments with questions.
Also for me it's unclear how to add size check in C code. I added method and definitions as in nullCheck, but I don't understand where this method should be called.

Would be appreciated if you help and/or point me to places using which I will be able to finish this PR ( https://github.com/konsoletyper/teavm/pull/868 depends on this feature).

Thanks in advance.


konsoletyper:
I don't think array size check should be performed by instruction transformation. Array access is guarded by a separate instruction that *can* be implemented efficiently by a particular backend. Introduction of basic blocks is too heavyweight and difficult in all possible aspects. However, a new instruction should be introduced with care and with clear understanding why it's done and how it will participate in all possible flows. In case of array construction, which is itself much more heavyweight, than array access, and which occurs much more rarely, introduction of new instructions or IR transformations don't worth it. I think corresponding check can be performed just within actual methods that construct array.

PR #872, opened 2023-12-04
Implement String.chars()

JFronny:
Simple implementation inspired by toCharArray.
Seems to work fine from my testing.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/stream/intimpl/TStringCharsStream.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/stream/intimpl/TStringCharsStream.java
@@ -25,7 +25,6 @@ public class TStringCharsStream extends TSimpleIntStreamImpl {
 
     public TStringCharsStream(TString string) {
         this.string = Objects.requireNonNull(string);
-        this.index = 0;
     }
 
     @Override
--- a/classlib/src/main/java/org/teavm/classlib/java/util/stream/intimpl/TStringCharsStream.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/stream/intimpl/TStringCharsStream.java
@@ -1,0 +1,49 @@
+/*
+ *  Copyright 2023 JFronny.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


JFronny:
What copyright owner should be specified in the license header of the stream class?

konsoletyper:
> What copyright owner should be specified in the license header of the stream class?

You can specify your own name. Or you can also implement stream with inner class.

JFronny:
> This PR lacks tests

I have implemented a simple test. How would I go about running it? The documentation mentions setup steps, but not how I would actually run the tests.

konsoletyper:
You should follow instructions in https://teavm.org/docs/contributing/testing.html and then run tests using your favourite way. You can just launch `./gradlew test` from console. Personally, I use IDEA for development and just press launch button across test method.

JFronny:
The issue seems to be that the test runner expects Google Chrome to be installed with a specific name and crashes when it can't find it. Using `-Pteavm.tests.browser=browser` and manually opening the browser works, though.

classlib/src/main/java/org/teavm/classlib/java/util/stream/intimpl/TStringCharsStream.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public TStringCharsStream(TString string) {
+        this.string = Objects.requireNonNull(string);
+    }
+
+    @Override
+    public boolean next(IntPredicate consumer) {
+        while (index < string.length()) {
+            if (!consumer.test(string.charAt(index++))) {
+                break;
```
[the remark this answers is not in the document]
JFronny: An equivalent alternative that avoids the unnecessary calculation but still checks after incrementing the index would be
```java
        while (index < string.length()) {
            if (!consumer.test(string.charAt(index++))) {
                return index < string.length();
            }
        }
        return false;
```
but I doubt that is preferred

classlib/src/main/java/org/teavm/classlib/java/util/stream/intimpl/TStringCharsStream.java
```diff
@@ -21,10 +21,11 @@
 
 public class TStringCharsStream extends TSimpleIntStreamImpl {
     private final TString string;
-    private int index = 0;
+    private int index;
 
     public TStringCharsStream(TString string) {
         this.string = Objects.requireNonNull(string);
+        this.index = 0;
```
konsoletyper, line 28: You don't need this. Java spec guarantees to initialize int fields with 0

JFronny: I know, but explicitly stating that we start at index 0 seemed nicer.

PR #873, opened 2023-12-04
Implement float support for String.format

### LINKED ISSUE #472: Float/Double Support for Formatter
Currently, something like `String.format("%f", 1.0f)` leads to TFormatter#formatValue:
`throw new UnknownFormatConversionException(String.valueOf(specifier));`

JFronny:
The actual implementation is entirely based on TDecimalFormat. The only extra thing is the calculation of the decimal size based on the target width of the string.
Additionally, this includes a fix for TDecimalFormat ignoring isGroupingUsed.
Probably closes #472

### DIFF
--- a/tests/src/test/java/org/teavm/classlib/java/text/DecimalFormatTest.java
+++ b/tests/src/test/java/org/teavm/classlib/java/text/DecimalFormatTest.java
@@ -489,8 +489,8 @@ public class DecimalFormatTest {
         format.setMaximumFractionDigits(6);
         format.setMinimumFractionDigits(6);
         format.setGroupingUsed(false);
-        assertEquals("1.200000", format.format((Object)1.2));
-        assertEquals("12.200000", format.format((Object)12.2));
+        assertEquals("1.200000", format.format((Object) 1.2));
+        assertEquals("12.200000", format.format((Object) 12.2));
         format.setMaximumFractionDigits(0);
         format.setMinimumFractionDigits(0);
         format.setMaximumIntegerDigits(5);
--- a/tests/src/test/java/org/teavm/classlib/java/util/FormatterTest.java
+++ b/tests/src/test/java/org/teavm/classlib/java/util/FormatterTest.java
@@ -265,13 +265,17 @@ public class FormatterTest {
         assertEquals("(1,234.6)", new Formatter(Locale.US).format("%0,(9.1f", -1234.6).toString());
 
         assertEquals("1 12 123 1,234 12,345 123,456 1,234,567", new Formatter(Locale.US)
-                .format("%,.0f %,.0f %,.0f %,.0f %,.0f %,.0f %,.0f", 1f, 12f, 123f, 1234f, 12345f, 123456f, 1234567f).toString());
+                .format("%,.0f %,.0f %,.0f %,.0f %,.0f %,.0f %,.0f", 1f, 12f, 123f, 1234f, 12345f, 123456f, 1234567f)
+                .toString());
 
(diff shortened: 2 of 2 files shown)


classlib/src/main/java/org/teavm/classlib/java/util/TFormatter.java
```diff
@@ -240,11 +241,67 @@ private void formatValue(char specifier) throws IOException {
...
+                format.setPositivePrefix("+"); // dfs has no plus sign
+            } else if ((flags & TFormattableFlags.LEADING_SPACE) != 0) {
+                format.setPositivePrefix(" ");
+            }
+
+            StringBuilder sb;
+            Object arg = args[argumentIndex];
+            if (arg instanceof Double) {
+                double value = (Double) arg;
+                sb = new StringBuilder(format.format(Math.abs(value)));
```
konsoletyper, line 276: What's the point in `Math.abs` here?

classlib/src/main/java/org/teavm/classlib/java/util/TFormatter.java
```diff
@@ -240,11 +241,67 @@ private void formatValue(char specifier) throws IOException {
...
+            } else if (arg instanceof Float) {
+                float value = (Float) arg;
+                sb = new StringBuilder(format.format(Math.abs(value)));
+            } else {
+                throw new IllegalFormatConversionException(specifier, arg != null ? arg.getClass() : null);
```
konsoletyper, line 281: BigDecimal should be also supported

classlib/src/main/java/org/teavm/classlib/java/util/TFormatter.java
JFronny: Because I forgot to remove it

classlib/src/main/java/org/teavm/classlib/java/util/TFormatter.java
```diff
@@ -240,11 +242,70 @@ private void formatValue(char specifier) throws IOException {
...
+            } else if (arg instanceof BigDecimal) {
+                BigDecimal value = (BigDecimal) arg;
+                sb = new StringBuilder(format.format(value));
+            } else {
+                throw new IllegalFormatConversionException(specifier, arg != null ? arg.getClass() : null);
+            }
+
+            if ((flags & TFormattableFlags.ZERO_PADDED) != 0) {
+                for (int i = sb.length(); i < width; ++i) {
+                    sb.append(Character.forDigit(0, 10));
```
konsoletyper, line 290: This should be zero digit from DecimalFormatSymbols. But anyway, I don't like this approach. I'd rather looked for a way to make TDecimalFormat do this padding

JFronny: This has been addressed with setMinimumIntegerDigits

JFronny:
`predictDecimalSize` is admittedly ugly, but since the width includes other characters and (to my knowledge) DecimalFormat provides no API that accepts a width, it seems needed.
Resetting the precision is also not great, but without that, the precision is either wrong or chars get trimmed by formatGivenString. If you want, I could append manually, though.

konsoletyper:
> predictDecimalSize is admittedly ugly, but since the width includes other characters and (to my knowledge) DecimalFormat provides no API that accepts a width, it seems needed.

Yes, that's why I think either TDecimalFormat could be extended compared to original DecimalFormat, or (which I believe a better solution) TDecimalFormat could be just a thin wrapper around internally implemented formatting library, as well as TFormatter. This also would solve the problem with locale-specific plus sign.

konsoletyper:
I'm ready to merge this PR, however it produces another error in WASI BE. This is not issue of this particular PR, but another issue in WASI BE, that did not uncover it yet. I'll merge this PR as soon as I fix issue in WASI BE.

PR #879, opened 2024-01-24
Fix/improve cloning of TreeMap and TreeSet objects

tryone144:
This PR fixes the shallow clone of `java.utils.TreeMap`. Previously, the clone kept the reference to the same root node and in turn the whole tree. Modifications to the clone incorrectly also modified the original TreeMap. This update generates new tree nodes while keeping the key and value references, correctly making the clone independent.

Additionally, this PR includes a `Clonable` implementation for `java.utils.TreeSet` that follows the same logic of `java.utils.HashSet`.


konsoletyper:
FYI: I encourage contributors to use `var` in their PR

PR #880, opened 2024-01-24
Extend BigInteger with xValueExact() and sqrt()

tryone144:
This PR improves the coverage of `java.math.BigInteger` by adding the missing `byteValueExact()`, `shortValueExact()`, `intValueExact()`, and `longValueExact()` methods, as well as an implementation for calculating the square root `sqrt()`.

In addition, the constant value `TWO` which has been added in Java 9, is added as well.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/math/TBigInteger.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/math/TBigInteger.java
@@ -68,6 +68,11 @@ public class TBigInteger extends Number implements Comparable<TBigInteger>, Seri
     public static final TBigInteger ONE = new TBigInteger(1, 1);
 
     /**
+     * The {@code BigInteger} constant 2.
+     */
+    public static final TBigInteger TWO = new TBigInteger(1, 2);
+
+    /**
      * The {@code BigInteger} constant 10.
      */
     public static final TBigInteger TEN = new TBigInteger(1, 10);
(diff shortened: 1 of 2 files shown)


classlib/src/main/java/org/teavm/classlib/java/math/TBigInteger.java
```diff
@@ -1097,6 +1146,76 @@ public TBigInteger pow(int exp) {
         return TMultiplication.pow(this, exp);
     }
 
+    /**
```
konsoletyper, line 1149: Just curious where you've taken this?

tryone144: The overall idea is taken from chapter 11 *Integer Square Root* using Newton's Method in the book [Hacker's Delight](https://doc.lagout.org/security/Hackers%20Delight.pdf), combined with some random code samples I've seen over the internet.

If I remember correctly, I got this reference from a doc-comment when scrolling through the OpenJDK classlib. Given that I have *seen* their code before writing this implementation, you might consider this implementation to be tainted by their GPL'ed code, even if not copied per-se.

If you prefer a "cleaner" implementation in this regard, I can provide an algorithmic write-up as a base for someone who hasn't seen their code.

tryone144: Yeah, I understand. This is fine by me.
Copyright and licensing is not always clear enough if you are not a lawyer and these instances are usually only clearly decided by a judge (with horrendous sums of money thrown at lawyers most people don't have). :shrug:

PR #884, opened 2024-02-01
JSO: add interface for Promise

tryone144:
This adds the `jso.core.JSPromise` interface to interact with native JavaScript `Promise`s.

I've added this to return Promises (and `async` computation results) back to the JavaScript context. Is there a built in way for asynchronous execution we can `await` on the JavaScript side? I like the `async/await` syntax with Promises more like completion callbacks (which are already possible in TeaVM).

<details>
<summary>Usage example:</summary>

I use this in my project by spawning a new thread and calling the `resolve` and `reject` methods after the inner thread has finished (or failed). The inner runner could be a simple function as well.

```java
import org.teavm.jso.core.JSFunction;
import org.teavm.jso.core.JSObjects;
import org.teavm.jso.core.JSPromise;
import org.teavm.jso.core.JSString;

/**
 * Thread executor that returns a JavaScript Promise which resolves/rejects depending on the execution result.
 */
static class ThreadExecutor extends Thread implements JSPromise.Executor {
    private Thread _thread;
    private int _timeout;

    private JSFunction _resolve;
    private JSFunction _reject;

    public ThreadExecutor(Thread thread, int timeout) {
        _thread = thread;
        _timeout = timeout;
    }

    /** Thread runner entry point */
    @Override
    public void run() {
        _thread.start();

        try {
            _thread.join();
        } catch (InterruptedException e) {
            e.printStackTrace();
            if (_reject != null && !JSObjects.isUndefined(_reject)) {
                _reject.call(this, JSString.valueOf("Interrupted"));
            }
        } catch (Exception e) {
            if (_reject != null && !JSObjects.isUndefined(_reject)) {
                _reject.call(this, JSString.valueOf(e.toString()));
            }
        } finally {
            if (_resolve != null && !JSObjects.isUndefined(_resolve)) {
                _resolve.call(this);
            }
        }
    }

    /** Promise runner entry point */
    @Override
    public void onExecute(JSFunction resolveFunc, JSFunction rejectFunc) {
        _resolve = resolveFunc;
        _reject = rejectFunc;

        this.start();
    }
}

/** Use it in a function */
public JSPromise execute(String code, JSNumber timeoutMs) {
    Thread runner = new Thread(() -> {}); // create Thread that does something with `code`
    int timeout = JSObjects.isUndefined(timeoutMs) ? 0 : timeoutMs.intValue();

    ThreadExecutor executor = new ThreadExecutor(runner, timeout);
    JSPromise promise = JSPromise.create(executor);

    return promise;
}
```

I think, this has quite a bit of boilerplate code, that would be the same for most use-cases. Maybe this can be abstracted away as well as part of the API?

</details>

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
+++ b/jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
@@ -1,0 +1,71 @@
+/*
+ *  Copyright 2023 Bernd Busse.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.jso.core;
+
+import org.teavm.interop.NoSideEffects;
+import org.teavm.jso.JSBody;
+import org.teavm.jso.JSByRef;
+import org.teavm.jso.JSFunctor;
+import org.teavm.jso.JSMethod;
+import org.teavm.jso.JSObject;
+
+public abstract class JSPromise implements JSObject {
+    private JSPromise() {
+    }
+
+    @JSFunctor
+    public interface Executor extends JSObject {
+        void onExecute(JSFunction resolveFunc, JSFunction rejectFunc);
+    }
+
+    @JSBody(params = "executor", script = "return new Promise(executor);")
+    @NoSideEffects
+    public static native JSPromise create(Executor executor);
+
+    @JSBody(params = "promises", script = "return Promise.any(promises);")
+    @NoSideEffects
+    public static native JSPromise any(@JSByRef JSArray promises);
+
+    @JSBody(params = "promises", script = "return Promise.all(promises);")
+    @NoSideEffects
+    public static native JSPromise all(@JSByRef JSArray promises);
+
+    @JSBody(params = "promises", script = "return Promise.allSettled(promises);")
+    @NoSideEffects
+    public static native JSPromise allSettled(@JSByRef JSArray promises);
+
+    @JSBody(params = "promises", script = "return Promise.race(promises);")
+    @NoSideEffects
+    public static native JSPromise race(@JSByRef JSArray promises);
+
+    @JSBody(params = "value", script = "return Promise.resolve(value);")
+    @NoSideEffects
+    public static native JSPromise resolve(JSObject value);
+
+    @JSBody(params = "reason", script = "return Promise.reject(reason);")
+    @NoSideEffects
+    public static native JSPromise reject(JSObject reason);
+
+    public abstract JSPromise then(JSFunction onFulfilled);
+
+    public abstract JSPromise then(JSFunction onFulfilled, JSFunction onRejected);
+
+    @JSMethod("catch")
+    public abstract JSPromise catch0(JSFunction onRejected);
+
+    @JSMethod("finally")
+    public abstract JSPromise finally0(JSFunction onFinally);
+}


jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
```diff
@@ -0,0 +1,71 @@
...
+package org.teavm.jso.core;
+
+import org.teavm.interop.NoSideEffects;
+import org.teavm.jso.JSBody;
+import org.teavm.jso.JSByRef;
+import org.teavm.jso.JSFunctor;
+import org.teavm.jso.JSMethod;
+import org.teavm.jso.JSObject;
+
+public abstract class JSPromise implements JSObject {
```
konsoletyper, line 25: I think this class would benefit from declaring it as generic.

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
```diff
@@ -0,0 +1,71 @@
...
+    private JSPromise() {
+    }
+
+    @JSFunctor
+    public interface Executor extends JSObject {
+        void onExecute(JSFunction resolveFunc, JSFunction rejectFunc);
```
konsoletyper, line 31: I'd use a more specific interface, like `JSConsumer<T>` for this purpose. I know, it's a bit inconvenient that `java.util.function` interfaces can't be used in this context. I'll consider to design support of these functional interfaces in JSO. For now you can declare something like `JSConsumer<T>` yourself and declare `resolveFunc` as `JSConsumer<T>`, and `rejectFunc` as `JSConsumer<Object>`

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
```diff
@@ -0,0 +1,71 @@
...
+    @JSBody(params = "reason", script = "return Promise.reject(reason);")
+    @NoSideEffects
+    public static native JSPromise reject(JSObject reason);
+
+    public abstract JSPromise then(JSFunction onFulfilled);
+
+    public abstract JSPromise then(JSFunction onFulfilled, JSFunction onRejected);
+
+    @JSMethod("catch")
+    public abstract JSPromise catch0(JSFunction onRejected);
```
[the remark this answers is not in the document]
tryone144: This was just uninspired copy from similar places in the classlib. `catchError` would match the implementation in `org.teavm.common.Promise`. I think I'll go with that.

What about the `finally0` below? The only other name I can come up with right now would be `finish`.

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
tryone144: Just to clarify whether I understand this correctly:
You'd prefer a JSO specific "copy" or extension of the `java.utils.function.Consumer` interface (for now), and use that as the type for the JavaScript supplied `resolveFunc` and `rejectFunc`? This is to be more specific about their signature (1 parameter, no return)?

Can't this `Executor` interface be a (custom) `BiConsumer` then, or does this still need more glue code in the code generation path?

What about the `then`, `catch` and `finally` methods? Is the `JSFunction` type correct for their arguments to pass Java functions to the JavaScript context, or would these also benefit from a more specific interface?

konsoletyper: You can replace `Executor` with a custom `BiConsumer`, but this results in a hard-to-understand signature like: `BiConsumer<Consumer<T>, Consumer<Object>>`, which is not quite readable and default parameter names won't be quite descriptive.

For `then`, `catch` and `finally` it's also necessary to introduce dedicated interfaces. `JSFunction` does not fit for these methods at all, since it's not a functional interface (i.e. does not provide exactly one method). It was designed for opposite case: *importing* JavaScript functions into Java (at least, the 'general' one, when developer is lazy enough to provide correct declaration with proper types and descriptive name). Please also note that `then` can return not only value itself, but Promise that resolves into value. In Java it's not possible to express due to lack of union types, so I suggest following: add another version of `then` method (e.g. `flatThen` or `thenPromise`) that expected to return promise. This won't solve *all* possible cases for JS promises like different branches of `if` inside lambda returning both value and promise, but this trade-off looks at least good enough.

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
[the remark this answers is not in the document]
konsoletyper: It can be `finish`, however, `finish` is a verb that describes what to do, whereas only few Promise methods are named with verbs, since then don't describe *what* to do but *how* to process resolved data. Thus they are all adverbs, like `then` or `finally`. So I suggest perhaps something descriptive like `finishedWith`, or `onFinish` or `onCompletion`, etc. Yes, my arguments seem a bit vague, since there's also `catch` method, which is a verb, so you can take is as advice in case you don't like `finish` yourself.

tryone144:
Adding generic type parameters made this class quite a bit more "complex". Is there a default location for JSO specific tests?

konsoletyper:
No, never even tried to test JSO APIs.

konsoletyper:
Don't see any purpose in that. With JSO you don't provide any implementations, just declarations. So what really need to be tested here is JSO itself (whether annotations propertly recognized). JSO declarations don't have regression, so never seen any point to write automatic tests.

tryone144:
> With JSO you don't provide any implementations, just declarations. …

Okay, makes sense. In this concrete case, this would mostly be to make sure my type parameters make sense. I'll push a WIP shortly.

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
```diff
@@ -0,0 +1,103 @@
...
+import org.teavm.jso.util.function.JSFunction;
+import org.teavm.jso.util.function.JSSupplier;
+
+/**
+ * Interface for interacting with JavaScript <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise">{@code Promise}s</a>.
+ *
+ * @param <T> The type this promise returns when resolving successfully.
+ * @param <E> The type this promise returns when rejecting.
+ */
+public abstract class JSPromise<T, E> implements JSObject {
```
konsoletyper, line 35: I think type parameter `E` is unnecessary. In TypeScript, `Promise` interface only declares type parameter for value type.

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
```diff
@@ -0,0 +1,103 @@
...
+        void onExecute(JSConsumer<T> resolveFunc, JSConsumer<E> rejectFunc);
+    }
+
+    @JSBody(params = "executor", script = "return new Promise(executor);")
+    @NoSideEffects
+    public static native <T, E> JSPromise<T, E> create(Executor<T, E> executor);
+
+    @JSBody(params = "promises", script = "return Promise.any(promises);")
+    @NoSideEffects
+    public static native JSPromise<?, ? extends JSError> any(@JSByRef JSArray promises);
```
konsoletyper, line 56: 1. This method should be generic
2. Instead of `JSArray<T>`, parameter should be `JSArrayReader<T>`. Actually, according to spec it must be `JSIterable<T>`, however, TeaVM does not support symbols and can't declare `JSIterable<T>`, so for now `JSArrayReader<T>` must be enough.

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @JSBody(params = "promises", script = "return Promise.all(promises);")
+    @NoSideEffects
+    public static native JSPromise<JSArray, ?> all(@JSByRef JSArray promises);
```
konsoletyper, line 60: There might be another version of `all` that takes something like `JSTuple2<T1, T2>, JSTuple3<T1, T2, T3>`, etc. However, these methods, as well as `JSTupleN` declarations, can be added after successful merge of this PR.

jso/apis/src/main/java/org/teavm/jso/util/function/JSConsumer.java
```diff
@@ -0,0 +1,32 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.jso.util.function;
+
+import java.util.Objects;
+
+@FunctionalInterface
+public interface JSConsumer<T> {
```
konsoletyper, line 21: I guess `@JSFunctor` annotation required here

jso/apis/src/main/java/org/teavm/jso/util/function/JSConsumer.java
```diff
@@ -0,0 +1,32 @@
...
+    void accept(T t);
+
+    default JSConsumer<T> andThen(JSConsumer<? super T> after) {
```
konsoletyper, line 24: Not sure JSO will handle this properly. Have you checked it?

tryone144: That's what my question about automated tests was about. :wink:
I'm in the process of writing some manual tests for the added functionality.

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
tryone144: Alright, I'll leave the `reject` case as a generic `Object` then.
I seem to remember a convention of rejecting with `Error` instances. But I don't think we can currently create `Error` instances directly.

jso/apis/src/main/java/org/teavm/jso/util/function/JSConsumer.java
tryone144: Yes, this annotation was required.
Somehow it wasn't properly working without this interface extending `JSObject` as well. Without it, passing lambda functions were rejected as "incompatible type".

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
tryone144: The `@JSByRef` annotation is also wrong...

jso/apis/src/main/java/org/teavm/jso/core/JSPromise.java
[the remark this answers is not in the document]
tryone144: I went with `catchError` and `flatCatchError` for a promises `catch()` method and `onSettled` for the `finally()` method.

konsoletyper:
Thanks! BTW, how is your project with TeaVM going?

tryone144:
> Thanks! BTW, how is your project with TeaVM going?

I have built a working prototype around july last year. Currently working on integrating the latest changes in TeaVM, polishing my changes to TeaVM and updating our build-system.

I will try replacing my hacky workaround for exporting Java functions and constants with the newly landed `@JSExport` annotations next. The new module generator is nicer than the one I had written. :smile: 

Be prepared for a PR regarding the `long`-emulation layer in the JS backend. I found some bottlenecks (at least on Firefox) with the use of `BigInt`s. Right now I am writing a benchmark sample to generate some baseline metrics in different environments.

konsoletyper:
> The new module generator is nicer than the one I had written.

Thanks. This is the very first prototype. I'm going to add various improvements eventually:

1. Allow to overload exported methods by signature, when possible
2. Allow to export constructors
3. Support varargs
4. Generate `d.ts` files
5. Verify exports and produce human-readable error messages

Anyway, I'm glad to get any feedback from users regarding this issue and possible improvements.

aghasemi:
Hi. Is this available in version 0.9.2?

konsoletyper:
@aghasemi no, it's available in `0.10.0-dev-8` and will be available in `0.10.0`

aghasemi:
Hi. I know the answer is very likely "It's done when it's done", but still, is there an estimate of when `0.10.0` will be out? Given how many Javascript APIs and libraries use Promises, this is a big change.

aghasemi:
Lack of expertise? :)

It seems non-intuitive to convert between a Java Future and a JS Promise, unlike the case of the data types. Can you give me some hints to start with?

Currently what I do as a workaround is to to write a call back in Javascript which does a `postMessage` and then listen to that message in TeaVM Java code.

konsoletyper:
> It seems non-intuitive for me to convert between a Java Future and a JS Promise, unlike the case of the data types. Can you give me some hints to start with?

Sorry, I don't fully understand what you want? Just JS promise declarations or some connection of `Future` interface? This PR addresses only former.

> Currently what I do as a workaround is to to write a call back in Javascript which does a postMessage and then listen to that message in TeaVM Java code.

How is it related to `Promise` class?

tryone144:
> It seems non-intuitive for me to convert between a Java Future and a JS Promise, unlike the case of the data types.

This PR only introduces an interface to interact with the [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) type in JavaScript. It shares its scope with the other JavaScript interop classes (`JSString`, `JSNumber`, `JSArray`, etc.) to allow interfacing with JavaScript methods that return `Promise`s or expect a `Promise` as a parameter.

This does neither provide nor intend to provide an asynchronous executor for Java code and the [`Future`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/util/concurrent/Future.html) interface. You might provide your own implementation using the `JSPromise` as a backend, though.

> Can you give me some hints to start with?

If you can't update to a 0.10 snapshot release, you can get away with copying this implementation into your project (minus the recent changes) and using this class directly.

If you really need an executor and the `Future` interface in the classlib, you have to provide your own implementation in your project and reference it in a [`META-INF/teavm.properties`](https://github.com/konsoletyper/teavm/blob/a1d711d0698eacd9bee8ab4a60fc6f69a84cd26a/classlib/src/main/resources/META-INF/teavm.properties#L17-L18) file. Easiest way would be to put these classes/interfaces in their own package and adjust the `map*` and `strip*` entries accordingly to have them appear in their proper place in the classlib hierarchy.

> Currently what I do as a workaround is to to write a call back in Javascript which does a postMessage and then listen to that message in TeaVM Java code.

What kind of problem are you trying to solve? Async execution in Java? :raised_eyebrow:

aghasemi:
> What kind of problem are you trying to solve? Async execution in Java? 🤨

Consider the following Javascript expression:

```javascript
import('https://cdn.jsdelivr.net/npm/@xenova/transformers').then(async ({pipeline}) => {
                return await pipeline('sentiment-analysis', 'Xenova/bert-base-multilingual-uncased-sentiment').then(async pipe => {
                    return await pipe('It may not be that bad, actually!').then(out => {
                        console.log(JSON.stringify(out))
                        return out
                    })
                })
            
            })
```

If you evaluate it, the result is Promise object, which when itself evaluated/executed, returns a JSON object (in this case `[{label=3 stars, score=0.33625108003616333}]`).


My question is, how to get to that JSON object in my TeaVM Java code, as a Future object or not.

konsoletyper:
@aghasemi you can get it as a `Promise` object. You don't necessarily need `Promise` declaration out-of-the box (honestly, the declarations in TeaVM are only for "quick start", and for real world scenarios you may need to declare your own bindings), but rather to write your own `Promise` "implementation". The key is that it's *super* easy, read [documentation](https://teavm.org/docs/runtime/jso.html)

aghasemi:
Thanks for the hint. For future reference, the following program correctly works for the code snippet mentioned above:

```java
package io.aghasemi;

import java.io.IOException;
import org.teavm.jso.JSBody;
import org.teavm.jso.JSFunctor;
import org.teavm.jso.json.JSON;


public class Client {
    
    private static final String jsCode =  """
        import('https://cdn.jsdelivr.net/npm/@xenova/transformers').then(async ({pipeline}) => {
            return await pipeline('sentiment-analysis', 'Xenova/bert-base-multilingual-uncased-sentiment').then(async pipe => {
                return await pipe('It may not be that bad, actually!').then(out => {
                    console.log(JSON.stringify(out))
                    return out
                })
            })
        
        }).then(result => {
            handler(result)
        })
    """;
    

    @JSFunctor
    public interface PromiseHandler extends JSObject {
        void then(JSObject result);
    }

    @JSBody(params = { "handler" }, script = jsCode)
static native void runPromiseCode(PromiseHandler handler);
    

    
    public static void main(String[] args) throws IOException {

        runPromiseCode(new PromiseHandler() {

            @Override
            public void then(JSObject result) {
                System.out.println("In Java code, the result is "+ JSON.stringify(result));
            }
            
        });
           
    }
}

```

Now, as the a rather minor issue, you see above that I had to `stringify` the returned JS object and likely later will have to again convert it to JSON in Java code to process it further. Can I directly get a usable (in Java) JSON object as `result`, or there is no way around it? 

Thanks again.

konsoletyper:
You can do with this `JSObject` whatever you want, it's just direct reference to real JavaScript object. For example, write a type declaration for this JSON object and cast it to required type. Or just declare this particular type

```java
void then(JSObject result);
```

instead of `JSObject`. You can also use something like `JSMapLike<JSObject>`, but then you'll end up with lots of casting and stringly typing. BTW, there's nothing mystical about `JSMapLike`, it's simply declared as

```javascript
public interface JSMapLike<T> extends JSObject {
    @JSIndexer
    T get(String key);

    @JSIndexer
    void set(String key, T object);
}
```

aghasemi:
Amazing. many thanks.

PR #888, opened 2024-02-14
JSO: make global objects available in WebWorker contexts

tryone144:
This removes the hardcoded `window` access in some globally available objects/function if they are also available in the WebWorker context (which has no `window` but `self`).

What do you think about changing these methods to instance-methods of `Window`? This would need a way to check between browser and worker context to select the proper function.
Alternatively, we could introduce a `teavm_globals` prefix to access the global scope in the `@JSBody` annotations if needed (like the runtime templates utilize).

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/browser/Window.java
+++ b/jso/apis/src/main/java/org/teavm/jso/browser/Window.java
@@ -190,9 +190,9 @@ public abstract class Window implements JSObject, WindowEventTarget, StorageProv
     @JSProperty
     public abstract double getDevicePixelRatio();
 
-    @JSBody(params = "s", script = "return window.atob(s);")
+    @JSBody(params = "s", script = "return atob(s);")
     public static native String atob(String s);
 
-    @JSBody(params = "s", script = "return window.btoa(s);")
+    @JSBody(params = "s", script = "return btoa(s);")
     public static native String btoa(String s);
 }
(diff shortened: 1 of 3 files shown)


konsoletyper:
> What do you think about changing these methods to instance-methods of Window? This would need a way to check between browser and worker context to select the proper function.

I don't think there's any difference between `atob` and `btoa` functions in different scopes.

konsoletyper:
> Alternatively, we could introduce a teavm_globals prefix to access the global scope in the @JSBody annotations if needed 

No, I don't see any need in this. TeaVM runtime has two distinct cases of accessing "global" scope: top-level runtime declarations and global declarations. `@JSBody` is not meant to access any of runtime functions (though there's a hack with hard-coded name prefix to do this, but it's a non-documented feature and a subject of change), so there's no need to introduce similar special. I think these `window.` prefixes were introduced here long ago, when TeaVM could not handle such cases properly.

tryone144:
> I don't think there's any difference between `atob` and `btoa` functions in different scopes.

As far as I am aware, these are identical. The differentiation would only be needed if we were to access these as instance methods of a Window object retrieved via `current()` or `worker()`. Judging by the other global methods, keeping these `static` is consistent.

> I think these `window.` prefixes were introduced here long ago, when TeaVM could not handle such cases properly.

If the identifiers in `@JSBody` are not subject to name mangling, this should be fine.

PR #899, opened 2024-03-27
Added files and streams API

Ihromant:


konsoletyper:
Please, remove input events. They aren't relevant to this PR

jso/apis/src/main/java/org/teavm/jso/file/JSBlob.java
```diff
@@ -0,0 +1,55 @@
...
+import org.teavm.jso.JSClass;
+import org.teavm.jso.JSObject;
+import org.teavm.jso.JSProperty;
+import org.teavm.jso.core.JSArray;
+import org.teavm.jso.core.JSPromise;
+import org.teavm.jso.streams.ReadableStream;
+import org.teavm.jso.typedarrays.ArrayBuffer;
+
+@JSClass(name = "Blob")
+public class JSBlob implements JSObject {
```
konsoletyper, line 27: What is the purpose of naming this `JSBlob` instead of `Blob`?

jso/apis/src/main/java/org/teavm/jso/file/JSFile.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.teavm.jso.file;
+
+import org.teavm.jso.JSClass;
+import org.teavm.jso.JSObject;
+import org.teavm.jso.JSProperty;
+import org.teavm.jso.core.JSArray;
+
+@JSClass(name = "File")
+public class JSFile extends JSBlob implements JSObject {
```
konsoletyper, line 24: What is the purpose of naming this `JSFile` instead of `File`?

jso/apis/src/main/java/org/teavm/jso/streams/ReadableStream.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public native JSPromise<?> abort(String reason);
+
+    public native JSPromise<?> cancel();
+
+    public native JSPromise<?> cancel(String reason);
+
+    public native JSPromise<?> pipeTo(WritableStream stream);
+
+    public native JSPromise<?> pipeTo(WritableStream stream, JSObject options);
```
konsoletyper, line 37: Please, add correct typing for `options`. I suggest something like this:

```java
public abstract class PipeOptions implements JSObject {
    @JSProperty
    private native void setPreventClose(boolean value);

    public PipeOptions preventClose(boolean value) {
        setPreventClose(value);
        return this;
    }

    // remaining properties

    @JSBody(script = "return {};")
    public static native PipeOptions create();
}

File: jso/apis/src/main/java/org/teavm/jso/file/JSBlob.java [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+    public JSBlob(JSArray<?> array) {
+
+    }
+
+    public JSBlob(JSArray<?> array, JSObject options) {
```
konsoletyper, line 32: Please, add declaration for options object

jso/apis/src/main/java/org/teavm/jso/file/JSFile.java
Ihromant: To not intersect with java.io.File

jso/apis/src/main/java/org/teavm/jso/file/JSBlob.java
Ihromant: To be consistent with JSFile

konsoletyper: You have more classes declared without JS prefix

jso/apis/src/main/java/org/teavm/jso/file/JSFile.java
konsoletyper: I don't think they will interfere. They are in different packages. Moreover, I can hardly imagine case when developer uses both `java.io.File` and `org.teavm.jso.file.File` within one source file.

PR #902, opened 2024-04-01
Input event

Ihromant:


Ihromant:
Hi @konsoletyper 
is something wrong with this PR that you're ignoring it?

Ihromant:
I added intermediate UIEvent because it is real intermediate JS interface between generic Event and specific events. Main purpose of this PR is to add InputEvent which I am using. Adding UIEvent can be avoided, but in this case getDetail should be added everywhere.

konsoletyper:
You should at least add corresponding `EventTarget` interface. And I still have no idea how exactly you are going to use it in any real-world scenario, at least in this form. Do you just detect the fact of a change in document, or you really check `inputType` property as well as `data` to extract something useful? In former case I still don't understand why you need a separate event. In the latter case how exactly you interpret `inputType`?

Ihromant:
I have large text area and checking what user is doing with its content before submitting. For this reason I need inputType (together with data) to track changes in this text area. I didn't find other way to achieve this without InputEvent.

Don't get the remark about eventTarget. It is inherited from Event interface, isn't it?

konsoletyper:
> I have large text area and checking what user is doing with its content before submitting.

So you basically ignore `inputType` and `data` properties, right?

> Don't get the remark about eventTarget. It is inherited from Event interface, isn't it?

Please, see how other events are implemented. For example, recent `TouchEvent`

jso/apis/src/main/java/org/teavm/jso/dom/events/InputEventTarget.java
```diff
@@ -0,0 +1,26 @@
...
+ *  limitations under the License.
+ */
+package org.teavm.jso.dom.events;
+
+public interface InputEventTarget extends EventTarget {
+    default void listenInput(EventListener<InputEvent> listener) {
+        addEventListener("input", listener);
+    }
+
+    default void neglectInput(EventListener<InputEvent> listener) {
```
konsoletyper, line 23: Please, remove this method

Ihromant:
> So you basically ignore inputType and data properties, right?
No, I'm using them.
> Please, see how other events are implemented. For example, recent TouchEvent
Done.

konsoletyper:
> > So you basically ignore inputType and data properties, right?
> 
> No, I'm using them.

How?

Ihromant:
> How?

For instance, I'm reacting differently on paste than on regular key input.

konsoletyper:
And please remove `UIEvent`. While other properties might be useful, your implementation does not provide them. As for `detail`, it's ugly designed in the standard and can be actually moved to `MouseEvent`, if you really need it.

PR #911, opened 2024-04-24
Add missing implementation of java.lang.Appendable interface to TeaVM java.io.PrintStream class

pcoperatr:
java.io.PrintStream is supposed to implement java.lang.Appendable (https://docs.oracle.com/javase/8/docs/api/java/io/PrintStream.html). But TeaVM's emulation is missing its implementation.

This fix solves the problem.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/io/TPrintStream.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/io/TPrintStream.java
@@ -18,8 +18,6 @@ package org.teavm.classlib.java.io;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Locale;
-import org.teavm.classlib.java.lang.TAppendable;
-import org.teavm.classlib.java.lang.TCharSequence;
 import org.teavm.classlib.java.lang.TMath;
 import org.teavm.classlib.java.lang.TObject;
 import org.teavm.classlib.java.lang.TStringBuilder;
@@ -33,7 +31,7 @@ import org.teavm.classlib.java.nio.charset.TUnsupportedCharsetException;
 import org.teavm.classlib.java.nio.charset.impl.TUTF8Charset;
 import org.teavm.classlib.java.util.TFormatter;
 
-public class TPrintStream extends TFilterOutputStream implements TAppendable {
+public class TPrintStream extends TFilterOutputStream implements Appendable {
     private boolean autoFlush;
     private boolean errorState;
     private TStringBuilder sb = new TStringBuilder();
@@ -271,34 +269,6 @@ public class TPrintStream extends TFilterOutputStream implements TAppendable {
         print('\n');
     }
 
-    @Override
-    public TPrintStream append(char c) {
-        print(c);
-        return this;
-    }
-
-    @Override
-    public TPrintStream append(TCharSequence csq) {
-        if (null == csq) {
-            sb.append("null");
-            printSB();
-        } else {
-            append(csq, 0, csq.length());
-        }
-        return this;
-    }
-
-    @Override
-    public TPrintStream append(TCharSequence csq, int start, int end) {
-        if (null == csq) {
-            sb.append("null");
-        } else {
-            sb.append(csq, start, end);
-        }
-        printSB();
-        return this;
-    }
-
     public TPrintStream printf(String format, Object... args) {
         return format(format, args);
     }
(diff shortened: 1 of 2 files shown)


classlib/src/main/java/org/teavm/classlib/java/io/TPrintStream.java
```diff
@@ -31,7 +33,7 @@
 import org.teavm.classlib.java.nio.charset.impl.TUTF8Charset;
 import org.teavm.classlib.java.util.TFormatter;
 
-public class TPrintStream extends TFilterOutputStream {
+public class TPrintStream extends TFilterOutputStream implements TAppendable {
```
konsoletyper, line 36: Take a look to the bottom of the file: there's a wrapper that create `Appendable` out of this `PrintStream`. I think that now `getAppendable` can be removed and formatters can take directly `this` reference

classlib/src/main/java/org/teavm/classlib/java/io/TPrintStream.java
```diff
@@ -269,6 +271,34 @@ public void println() {
...
+    }
+
+    @Override
+    public TPrintStream append(TCharSequence csq, int start, int end) {
+        if (null == csq) {
+            sb.append("null");
+        } else {
+            sb.append(csq, start, end);
+        }
+        printSB();
```
konsoletyper, line 298: perhaps you can use `print()` instead and avoid StringBuilder

classlib/src/main/java/org/teavm/classlib/java/io/TPrintStream.java
pcoperatr: I made TPrintStream implement TAppendable, and not java.lang.Appendable interface. So we can't get rid of getAppendable() unless TPrintStream implements Appendable.

I didn't find guidelines for design of classlib classes, so intuitively decided that class with the name starting with T should implement T interfaces.

If you are OK to make TPrintStream implement java.lang.Appendable instead of TAppendable I can alter the fix.

konsoletyper: TeaVM does following:

1. As a class required and it's in a package starting with `java.`, adds `org.teavm.classlib.` prefix to the package name and `T` prefix to the class name and takes this class, then renames it back to initial name.
2. After any class is parsed, all occurrences of classes (in method parameters, invocation instructions, etc) are renamed if they are in package `org.teavm.classlib.java.` using previous pattern.

So it does not matter if you implement `Appendable` or `TAppendable`.

pcoperatr: Updated the fix as agreed. TPrintStream looks cleaner now

classlib/src/main/java/org/teavm/classlib/java/io/TPrintStream.java
```diff
@@ -301,30 +301,21 @@ private void printSB() {
...
-        return appendable;
+    @Override
+    public Appendable append(CharSequence csq) {
+        print(csq, 0, csq.length());
+        return this;
+    }
+
+    @Override
+    public Appendable append(CharSequence csq, int start, int end) {
+        print(csq, start, end);
```
konsoletyper, line 312: I guess you should change first parameter to something like this: `csq != null ? csq : "null"`. Otherwise tests fail.

pcoperatr: Right, my bad. Fixed this one.

BTW, how can I run a single TeaVM test? What should be the command line options for this? I don't want to run the full test set, but PrintStreamTest class, or even a single PrintStreamTest#append_null method.

konsoletyper: > BTW, how can I run a single TeaVM test?

Just like any other single JUnit test. For example [see this answer](https://stackoverflow.com/questions/22505533/how-to-run-only-one-unit-test-class-using-gradle). Or just get IDEA and click arrow button right behind the method line.

pcoperatr: I use NetBeans and it can run individual tests for Maven projects, but not for Gradle projects. At least I don't have luck with this IDE. Thank you for the link!

konsoletyper:
Tests still failing. One reason is that `append` methods have wrong signature - they must return `PrintStream` instead of `Appendable`. Then tests go a little bit further but fail for some reason - I'll investigate

pcoperatr:
I can reproduce the failure. Will work on the fix

pcoperatr:
The last commit fixes all test case failures, but now the test hangs. It behaves similar to https://github.com/konsoletyper/teavm/issues/913 issue

Issue #916, opened 2024-05-03
Regression with svg elements


Ihromant:
After some recent changes new issue appeared:
```java
HTMLElement svg = HTMLDocument.current().createElementNS("http://www.w3.org/2000/svg", "svg").cast();
        svg.setAttribute("width", "10");
        HTMLDocument.current().appendChild(svg);
```
fails with CCE. It compiles to
```javascript
var$2 = otj_JSObject_cast$static((otjdh_HTMLDocument_current()).createElementNS("http://www.w3.org/2000/svg", "svg"));
    var$3 = $rt_instanceOfOrNull(var$2, HTMLElement);
    $svg = $rt_throwCCEIfFalse(var$3, var$2);
    $svg.setAttribute("width", "10");
    (otjdh_HTMLDocument_current()).appendChild($svg);
```
It fails because svg is not instance of HTMLElement.
I overcame it with 
```java
@JSBody(script = "return document.createElementNS('http://www.w3.org/2000/svg', 'svg');")
    private static native HTMLElement svg();
```
but earlier it worked well.

konsoletyper:
This is expected behaviour. Unfortunately, previous versions of TeaVM did not behave correctly with run-time type checking in case of JSO, so you did not identify this issue on your side. The proper way would be to upcast `HTMLDocument` to just `Document` and then use it to create SVG elements. I know, standard does not re-define return value of `createElement` to return `HTMLElement`, but it looked convenient and assuming most users would work with HTML DOM and removing override to `createElement` is a breaking change, I won't remove it.

PR #917, opened 2024-05-03
Element fix

Ihromant:
https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML
InnerHTML and other properties are a properties of Element, but not HTMLElement. This API mismatch caused inproper usage of JSO api which caused working incorrect code https://github.com/konsoletyper/teavm/issues/916 
Incorrect code is fixed now, so properties should be moved to correct place.

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/dom/html/HTMLElement.java
+++ b/jso/apis/src/main/java/org/teavm/jso/dom/html/HTMLElement.java
@@ -135,12 +135,6 @@ public abstract class HTMLElement implements Element, ElementCSSInlineStyle, Eve
     public abstract HTMLCollection getChildren();
 
     @JSProperty
-    public abstract String getInnerHTML();
-
-    @JSProperty
-    public abstract void setInnerHTML(String content);
-
-    @JSProperty
     public abstract String getInnerText();
 
     @JSProperty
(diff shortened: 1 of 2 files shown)


Ihromant:
Why closed this PR? It isn't incorrect and puts properties to the correct place and is backward compatible.

konsoletyper:
This PR is incorrect.

Ihromant:
It actually fixes issues I presented in https://github.com/konsoletyper/teavm/issues/916 when I create ElementNS, I can assign class to its classlist or setInnerHTML in JS. Due to this methods presence in JSO HTMLElement, this methods are inaccessible at all. What are you suggesting then?

konsoletyper:
I don't understand the problem. Can you please clarify?

Ihromant:
```java
Element svg = HTMLDocument.current().createElementNS("svgschema", "svg);
// i want to set inner HTML or classList but I can't because it's not present in Element interface
```
Earlier it worked due to wrong typechecking on JS side. Due to this fact, I didn't check that it's not in Element and used it against HTMLElement. Now it is broken because you fixed JS type system. So now we need to put properties in correct place. Please check MDN link I provided.

konsoletyper:
First of all, MDN is non-normative and you can't refer on it. Second, there are two `Element` interfaces: from DOM spec and from HTML spec, and the one from DOM spec does not contain `innerHTML` property, and especially properties like `clientWidth`, since in general case a DOM application does not provide any visual representation.

Now, I just wasted 10 minutes of my time to explain you that instead of just improving TeaVM. When I said "this PR is incorrect" this means this PR is incorrect and you should accept it and try to figure out why youself.

PR #931, opened 2024-07-03
JSO contribution of TouchEvent Wrapper.

pizzadox9999:
Hi, I would like to contribute the TouchEvent wrappers. If the wrappers need any adjustment please let me know. Best regards pizzadox9999

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/dom/events/Touch.java
+++ b/jso/apis/src/main/java/org/teavm/jso/dom/events/Touch.java
@@ -1,0 +1,101 @@
+package org.teavm.jso.dom.events;
+
+import org.teavm.jso.JSObject;
+import org.teavm.jso.JSProperty;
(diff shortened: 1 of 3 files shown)


konsoletyper:
Please, look at the `onSomeEvent` convention used elsewhere. For example, [MouseEventTarget](https://github.com/konsoletyper/teavm/blob/efcb22b639448fe558d271aa80417d3317c88a50/jso/apis/src/main/java/org/teavm/jso/dom/events/MouseEventTarget.java)

pizzadox9999:
Sorry, for the long time. I have a question about your request. Do you mean I should create an EventTarget like the MouseEventTarget? Something else didn't come to my mind. If it isn't the EventTarget I need to write, would you please elaborate a bit about what else should be done. But some events don't have an EventTarget? Thank you pizzadox9999

PR #949, opened 2024-09-07
jso: fix JSSymbol implementation

skerit:
I think some test code made it into the `JSSymbol` class in commit a6fb67817c5bd57c19df2ed1e18fd9831a9d30d6, because suddenly it lost its `abstract` status, and its method had the wrong annotations, causing projects to no longer compile when using a JSSymbol instance.

This basically restores that, and adds some basic tests.

I don't know if you still want to use the `@NoSideEffects` annotation though.

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/core/JSSymbol.java
+++ b/jso/apis/src/main/java/org/teavm/jso/core/JSSymbol.java
@@ -15,24 +15,24 @@
  */
 package org.teavm.jso.core;
 
-import org.teavm.interop.NoSideEffects;
 import org.teavm.jso.JSBody;
-import org.teavm.jso.JSIndexer;
 import org.teavm.jso.JSObject;
 import org.teavm.jso.JSPrimitiveType;
 
 @JSPrimitiveType("symbol")
-public class JSSymbol<T> implements JSObject {
+public abstract class JSSymbol<T> implements JSObject {
     private JSSymbol() {
     }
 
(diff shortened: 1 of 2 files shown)


jso/apis/src/main/java/org/teavm/jso/core/JSSymbol.java
```diff
@@ -15,24 +15,24 @@
...
 
-import org.teavm.interop.NoSideEffects;
 import org.teavm.jso.JSBody;
-import org.teavm.jso.JSIndexer;
 import org.teavm.jso.JSObject;
 import org.teavm.jso.JSPrimitiveType;
 
 @JSPrimitiveType("symbol")
-public class JSSymbol<T> implements JSObject {
+public abstract class JSSymbol<T> implements JSObject {
```
konsoletyper, line 23: Please, remove `abstract` modifier and add `@JSClass` annotation instead

tests/src/test/java/org/teavm/jso/test/JSWrapperTest.java
```diff
@@ -331,6 +332,28 @@ public void jsToString() {
         assertEquals("bar", b.toString());
     }
 
+    @Test
+    public void jsSymbol() {
```
konsoletyper, line 336: `JSWrapperTest` has nothing to do with `JSSymbol`. Consider creating a separate test class for `JSSymbol`

PR #950, opened 2024-09-09
Small BitSet fixes and improvements

Ihromant:


classlib/src/main/java/org/teavm/classlib/java/util/TBitSet.java
```diff
@@ -76,12 +86,19 @@ public static TBitSet valueOf(byte[] bytes) {
...
+        buff = buff.slice().order(TByteOrder.LITTLE_ENDIAN);
+        byte[] words = new byte[buff.remaining()];
+        buff.get(words);
+        return valueOf(words);
+    }
+
     public byte[] toByteArray() {
         byte[] bytes = new byte[(length + 7) / 8];
-        int fullInts = length / TInteger.SIZE;
+        int fullInts = bytes.length / 4;
```
konsoletyper, line 98: Why did you replace `TInteger.SIZE` with `4`?

classlib/src/main/java/org/teavm/classlib/java/util/TBitSet.java
```diff
@@ -76,12 +86,19 @@ public static TBitSet valueOf(byte[] bytes) {
...
 
+    public static TBitSet valueOf(TByteBuffer buff) {
+        buff = buff.slice().order(TByteOrder.LITTLE_ENDIAN);
+        byte[] words = new byte[buff.remaining()];
+        buff.get(words);
+        return valueOf(words);
+    }
+
     public byte[] toByteArray() {
         byte[] bytes = new byte[(length + 7) / 8];
```
konsoletyper, line 97: I think there was an issue here. It should be `((length + 3) / 4)` (or better `((length + TInteger.SIZE - 1) / TInteger.SIZE)`

konsoletyper: Ah, no, it's ok

classlib/src/main/java/org/teavm/classlib/java/util/TBitSet.java
Ihromant: In case when you have length = 25, you have an issue. THis change removes the issue. It should be either (length + 7) / 32 or bytes.length / 4. I used second because we already calculated bytes.length.

konsoletyper, line 98: Why not `TInteger.BYTES` then?

PR #956, opened 2024-10-02
Send beacon

Ihromant:
Hi @konsoletyper . This PR adds navigator.sendBeacon function. Still, when I looked at navigator class, I noticed that all its methods are static. This is very different from Window or Document, which are abstract, but are instantiated using Window.current() or HTMLDocument.current(). Is it made for a reason and is it working or it is just a mistake that happened during the refactoring?


konsoletyper:
> This is very different from Window or Document, which are abstract, but are instantiated using Window.current() or HTMLDocument.current(). Is it made for a reason and is it working or it is just a mistake that happened during the refactoring?

It was made by reason. There can be multiple instances of Document or Window per app, for example, when parsing XML/HTML manually (and thus using DOM to navigated detached document) or when using iframes. However, Navigator only needs single instance per browser. If you look at its methods, it becomes quite clear that it does not act in the context of a frame or a tab.

Issue #971, opened 2024-11-17
WASM GC: Avoid using `array.new_fixed` to allocate arrays over 10000 elements


lax1dude:
I got my app compiled to WASM GC with my own WASM heap direct buffer implementation, but Chrome complains when I try loading the file via `WebAssembly.compileStreaming` that it’s attempting to create a 17626 element array using `array.new_fixed` and that the limit for creating arrays like this is 10000 elements.

`Uncaught (in promise) CompileError: WebAssembly.compileStreaming(): Compiling function #37428 failed: Requested length 17626 for array.new_fixed too large, maximum is 10000 @+6497484`

Turns out its because the string pool and some other initialization code uses it, and my app is so absolutely massive that is has too many strings. I was able to fix this on a new branch in my fork and then actually compile my app to a valid WASM binary that can be loaded in Chrome, however I'm not spend time making a pull request for this unless you are actually interested. 

[https://github.com/konsoletyper/teavm/compare/master...lax1dude:teavm:array-new-fixed-patch](https://github.com/konsoletyper/teavm/compare/master...lax1dude:teavm:array-new-fixed-patch)

konsoletyper:
I think this patch is excessive. In normal code there's only small chance the we ever reach this limit (we would rather reach method instruction count limit before that). So the only practically possible case is string pool initialization. For that case the solution is simple: just cut entire string list into chunks of at most 10000 elements and pass each one to `teavm@initStrings`.

lax1dude:
I see, that would probably also make a smaller binary. I guess I'll eventually redo it when I have time. I think the reason I got my mind set on just replacing all usages of `WasmArrayNewFixed` was because I wasn't sure if the string pool was the only array that was too big.

PR #975, opened 2024-11-28
WASM GC: Fix backwards array copy function

lax1dude:
The source and target offset parameters in the `System.arraycopy` implementation are passed to `array.copy` in the wrong order

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/generate/gc/classes/WasmGCClassGenerator.java
+++ b/core/src/main/java/org/teavm/backend/wasm/generate/gc/classes/WasmGCClassGenerator.java
@@ -876,8 +876,8 @@ public class WasmGCClassGenerator implements WasmGCClassInfoProvider, WasmGCInit
         var targetArrayData = new WasmStructGet(arrayStruct, targetArray, ARRAY_DATA_FIELD_OFFSET);
 
         function.getBody().add(new WasmArrayCopy(
-                arrayDataType, targetArrayData, new WasmGetLocal(sourceIndexLocal),
-                arrayDataType, sourceArrayData, new WasmGetLocal(targetIndexLocal),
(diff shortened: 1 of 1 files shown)


lax1dude:
I don't know why the PR unit test failed, the unit tests that ran when I pushed the commit 30 minutes before passed, and the branch I made was based on your latest commit, I think the failure is unrelated.

[https://github.com/lax1dude/teavm/commit/75ab3beb131a33b415d8a7de0dd76cd32db50127](https://github.com/lax1dude/teavm/commit/75ab3beb131a33b415d8a7de0dd76cd32db50127)

lax1dude:
> Seems it's a false positive. For some reason tests sometimes fail on github, however never could reproduce this locally. Can you please provide tests for this fix?

Regarding if the patch works or not, my app wasn't working before and was crashing immediately due to an out of bounds array access trap on a copy, and when I swapped those two parameters it worked perfectly and now it the app loads all the way to the title screen with no issues.

Are you asking me to go and revise your existing classlib unit tests so that arraycopy is tested with separate src/dst offset values? If that's the case I can take care of that sometime when I also clean up my fix for the string pool being too large based on your recommendation from #971.

konsoletyper:
> Are you asking me to go and revise your existing classlib unit tests so that arraycopy is tested with separate src/dst offset values

Yes, exactly. This should be super-easy, why don't you do it right now? It should take 15 minutes.

konsoletyper:
Ok, I added tests myself. And it took 5 minutes to write them. Writing tests is an important part of software engineering culture, so please, if you submit PR and there's no reason *not* to write tests (like writing good test would take hours), you should add tests.

lax1dude:
> Ok, I added tests myself. And it took 5 minutes to write them. Writing tests is an important part of software engineering culture, so please, if you submit PR and there's no reason not to write tests (like writing good test would take hours), you should add tests.

Okay

PR #977, opened 2024-11-30
WASM GC: Add support for string pools over 10,000 strings

### LINKED ISSUE #971: WASM GC: Avoid using `array.new_fixed` to allocate arrays over 10000 elements

lax1dude:
Fixes #971, I've tested this on my app with 17000 strings and it works.

Would you like it if I also added a unit test with 10,001 string constants in it to test this, or would that be excessive?

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/generate/gc/strings/WasmGCStringPool.java
+++ b/core/src/main/java/org/teavm/backend/wasm/generate/gc/strings/WasmGCStringPool.java
@@ -93,11 +93,17 @@ public class WasmGCStringPool implements WasmGCStringProvider, WasmGCInitializer
                     void.class));
             function.getBody().add(new WasmCall(internInit));
         }
-        var array = new WasmArrayNewFixed(stringsArray);
-        for (var str : stringMap.values()) {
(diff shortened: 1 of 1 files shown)


konsoletyper:
Sure. This is the case when testing is too complex

PR #978, opened 2024-11-30
WASM: Intrinsify numberOfLeadingZeros, numberOfTrailingZeros, and bitCount

lax1dude:
I implemented intrinsic functions for `numberOfLeadingZeros`, `numberOfTrailingZeros`, and `bitCount` using the `clz`, `ctz`, and `popcnt` instructions in the WASM and WASM GC backends.

The classlib unit tests pass and my app still works, I make use of the leading and trailing zero functions in my direct buffer malloc/free implementation so I actually do have a reason to optimize them.

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/WasmTarget.java
+++ b/core/src/main/java/org/teavm/backend/wasm/WasmTarget.java
@@ -115,6 +115,7 @@ import org.teavm.backend.wasm.render.WasmBinaryWriter;
 import org.teavm.backend.wasm.render.WasmCRenderer;
 import org.teavm.backend.wasm.render.WasmRenderer;
 import org.teavm.backend.wasm.runtime.WasmSupport;
+import org.teavm.backend.wasm.transformation.BaseClassesTransformation;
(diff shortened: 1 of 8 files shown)


konsoletyper:
I doubt it will work with optimizations enabled. Optimizer *may* inline method bodies, written in Java, which makes impossible to apply these intrinsics. So additionally Wasm GC BE should provide class transformers which turns intrinsic target methods into native and removes their body.

lax1dude:
@konsoletyper I see what you mean, I'll fix it tomorrow

konsoletyper:
Yes, and although I won't ask you to write unit tests, please, at least disassemble your app and make sure that relevant WebAssembly instructions are used instead of falling back to "software" implementations.

lax1dude:
@konsoletyper Okay I added the class transformers. I also added unit tests for the 64-bit `numberOfLeadingZeros` and `numberOfTrailingZeros` because I noticed they didn't exist. I also copied the code from the unit tests and compiled it myself and disassembled the binary and all the intrinsics are compiling correctly (on the `AGGRESSIVE` optimization level).

PR #998, opened 2025-01-20
WASI VirtualSocket interface and Java networking classes implementation

mandesero:
This pull request introduces two key updates to enhance networking capabilities in a WASI-based environment:  

### Patch 1: **WASI: VirtualSocket Interface**  
This patch provides the core functionality required for socket operations within the WASI environment. Key features include:  
- Creating, binding, connecting, and listening to sockets.  
- Sending and receiving data.  
- Configuring socket options such as broadcast and reuse.  
- Retrieving peer and socket names.  

⚠️ This interface is specific to WASI-based WebAssembly and is not supported for JavaScript or non-WASI WebAssembly environments.  

### Patch 2: **classlib: Implement java.net.Socket**  
This patch introduces the implementation of the java.net.Socket class, enabling developers to work with sockets in a WASI-based environment. Additionally, the following related networking classes are implemented:  
- java.net.ServerSocket
- java.net.InetAddress  
- java.net.Inet4Address  
- java.net.Inet6Address  
- java.net.SocketAddress  
- java.net.InetSocketAddress  
- java.net.Proxy 
- java.net.NetworkInterface  
- java.net.SocketException  
- java.net.UnknownHostException

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/net/TSocket.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/net/TSocket.java
@@ -1,0 +1,521 @@
+/*
+ *  Copyright 2025 Maksim Tiushev.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.classlib.java.net;
+
+import static org.teavm.backend.wasm.wasi.Wasi.INET4;
+import static org.teavm.backend.wasm.wasi.Wasi.INET6;
+import static org.teavm.backend.wasm.wasi.Wasi.INET_UNSPEC;
+import static org.teavm.backend.wasm.wasi.Wasi.SHUT_RD;
+import static org.teavm.backend.wasm.wasi.Wasi.SHUT_RDWR;
+import static org.teavm.backend.wasm.wasi.Wasi.SHUT_WR;
+import static org.teavm.backend.wasm.wasi.Wasi.SOCK_DGRAM;
+import static org.teavm.backend.wasm.wasi.Wasi.SOCK_STREAM;
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import org.teavm.backend.wasm.runtime.net.impl.SockAddrInet4;
+import org.teavm.backend.wasm.runtime.net.impl.SockAddrInet6;
+import org.teavm.runtime.net.SockAddr;
+import org.teavm.runtime.net.VirtualSocket;
+import org.teavm.runtime.net.VirtualSocketProvider;
+
+public class TSocket implements Closeable {
+    protected int fd;
+    protected boolean created;
+
+    private static final int FD_CLOSED = -1;
+
+    private static VirtualSocket vs() {
+        return VirtualSocketProvider.getInstance();
+    }
+
+    public TSocket(int fd) {
+        this.fd = fd;
+        this.created = true;
+    }
+
+    public int getFd() {
+        return fd;
+    }
+
+    public TSocket() throws IOException {
+        this.fd = vs().socket(INET_UNSPEC, SOCK_STREAM);
+        this.created = true;
+    }
+
+    public TSocket(TInetAddress address, int port) throws IOException {
+        this(address, port, true);
+    }
+
+    public TSocket(TInetAddress host, int port, boolean stream) throws IOException {
+        this(host != null ? new TInetSocketAddress(host, port) : null,
+            (TSocketAddress) null, stream);
+    }
+
+    public TSocket(TInetAddress address, int port, TInetAddress localAddr, int localPort)
+            throws IOException {
+        this(address != null ? new TInetSocketAddress(address, port) : null,
+            new TInetSocketAddress(localAddr, localPort), true);
+    }
+
+    public TSocket(TProxy proxy) throws IOException {
+        throw new UnsupportedOperationException();
+    }
+
+    public TSocket(String host, int port) throws IOException {
+        this(host, port, true);
+    }
+
+    public TSocket(String host, int port, boolean stream) throws IOException {
+        this(host != null ? new TInetSocketAddress(host, port)
+                : new TInetSocketAddress(TInetAddress.getByName(null), port),
+                (TSocketAddress) null, stream);
+    }
+
+    public TSocket(String host, int port, TInetAddress localAddr, int localPort)
+            throws IOException {
+        this(host != null ? new TInetSocketAddress(host, port)
+                : new TInetSocketAddress(TInetAddress.getByName(null), port),
+                new TInetSocketAddress(localAddr, localPort), true);
+    }
+
+    private TSocket(TSocketAddress addr, TSocketAddress localAddr, boolean stream)
+            throws IOException {
+        assert addr instanceof TInetSocketAddress;
+        int sotype = stream ? SOCK_STREAM : SOCK_DGRAM;
+
+        SockAddr sa = getSockAddr(addr);
+        int family = sa.sockFamily();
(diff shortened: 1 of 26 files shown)


konsoletyper:
First of all, did you get any code from OpenJDK?

konsoletyper:
I have some ideas how to test actual sockets. However, something is missing in current test framework, so I won't force you to write tests for Socket

mandesero:
> First of all, did you get any code from OpenJDK?


I reviewed the OpenJDK sources, and only small classes that do not require specific logic, such as [SocketAddress](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/net/SocketAddress.java), [SocketException](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/net/SocketException.java), and [UnknownHostException](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/net/UnknownHostException.java), may coincide. All other code was written entirely by me.

konsoletyper:
You could have gotten sources from Harmony project

konsoletyper:
How to test this PR? I tried both wasmer and wasmtime and they both complain about missing `wasi_snapshot_preview1::sock_open` import. Do you know any runtime that supports this?

core/src/main/java/org/teavm/backend/wasm/runtime/net/WasiVirtualSocket.java
```diff
@@ -0,0 +1,428 @@
...
+        }
+        return dataLen[0];
+    }
+
+    @Override
+    public int recvFrom(int fd, byte[] buf, int len, SockAddr sa) throws SocketException {
+        byte[] addr = new byte[ADDR_SIZE];
+        Address rawAddr = Address.ofData(addr);
+
+        ByteBuffer buffer = ByteBuffer.allocate(BUFFER_SIZE);
```
konsoletyper, line 123: You could use structures for this purpose. First of all, instead of `BUFFER_SIZE`, you can use `Structure.sizeOf`, and you don't need buffers

classlib/src/main/java/org/teavm/classlib/java/net/TSocket.java
```diff
@@ -0,0 +1,521 @@
...
+    public InputStream getInputStream() throws IOException {
+        ensureSocketOpen();
+        final SockAddr remoteAddress = vs().getPeerName(fd);
+
+        return new InputStream() {
+            private static final int MAX_LENGTH = 8192;
+            private final byte[] buffer = new byte[MAX_LENGTH];
+
+            @Override
+            public int read() throws IOException {
```
konsoletyper, line 236: it would be also good to implement `available` method

core/src/main/java/org/teavm/backend/wasm/runtime/net/impl/SockAddrInet6.java
```diff
@@ -0,0 +1,140 @@
...
+            throw new IllegalArgumentException("Port number out of range: " + port);
+        }
+        return port;
+    }
+
+    private static String[] parseAddrPort(String addrPort) {
+        if (!addrPort.contains("[") || !addrPort.contains("]")) {
+            throw new IllegalArgumentException("Invalid format. Expected format: '[ipv6]:port'.");
+        }
+        String[] parts = addrPort.split("]:\");");
```
konsoletyper, line 75: It is a syntax error?

core/src/main/java/org/teavm/backend/wasm/runtime/net/impl/SockAddrInet6.java
```diff
@@ -0,0 +1,140 @@
...
+        if (parts.length != 2) {
+            throw new IllegalArgumentException("Invalid format. Expected format: '[ipv6]:port'.");
+        }
+        return new String[] {parts[0].substring(1), parts[1]};
+    }
+
+    private static short[] parseIPv6(String ip) {
+        String[] segments = ip.split(":");
```
konsoletyper, line 83: Please, don't use `String.split`!

core/src/main/java/org/teavm/backend/wasm/runtime/net/WasiVirtualSocket.java
konsoletyper: BTW, for passing small structures you can use `WasiBuffer.getBuffer()`

konsoletyper (changes requested):

mandesero:
> How to test this PR? I tried both wasmer and wasmtime and they both complain about missing `wasi_snapshot_preview1::sock_open` import. Do you know any runtime that supports this?

Sorry for the long delay. You should use [WAMR (WebAssembly Micro Runtime)](https://github.com/bytecodealliance/wasm-micro-runtime), which provides support for WASI socket APIs. You can run your `.wasm` file using [iwasm](https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/product-mini)

###  **Building `iwasm` with WASI socket support**
Ensure you have all dependencies installed before proceeding [(see)](https://github.com/bytecodealliance/wasm-micro-runtime/tree/main/product-mini#linux).

```bash
cd product-mini/platforms/linux
mkdir build && cd build

# Configure the build with necessary features
cmake .. \
  -DWAMR_BUILD_PLATFORM=linux \
  -DWAMR_BUILD_TARGET=X86_64 \
  -DWAMR_BUILD_INTERP=1 \
  -DWAMR_BUILD_FAST_INTERP=1 \
  -DWAMR_BUILD_AOT=1 \
  -DWAMR_BUILD_LIBC_BUILTIN=1 \
  -DWAMR_BUILD_LIBC_WASI=1 \
  -DWAMR_BUILD_THREAD_MGR=1 \
  -DWAMR_BUILD_SHARED_MEMORY=1 \
  -DWAMR_BUILD_SIMD=1

make
# The 'iwasm' binary will be generated in the current directory
```

### **Running your WASM file with `iwasm`**
Once built, you can execute your `.wasm` module using `iwasm`:

```bash
./iwasm --addr-pool=0.0.0.0/0 --allow-resolve="*" YourWasmFile.wasm
```

classlib/src/main/java/org/teavm/classlib/java/net/TSocket.java
mandesero: Done.

konsoletyper:
When you make fixes in PR, please, don't force-push rebased commit. Instead, you should make commits that fix original commits. When I merge your changes, I'll rebase myself. Otherwise it's extremely hard to track what has changed!

core/src/main/java/org/teavm/backend/wasm/runtime/net/WasiVirtualSocket.java
mandesero: I've rewritten the classes to use structures as recommended.

mandesero:
> When you make fixes in PR, please, don't force-push rebased commit. Instead, you should make commits that fix original commits. When I merge your changes, I'll rebase myself. Otherwise it's extremely hard to track what has changed!

Sorry, I'll make sure to follow this approach next time.

core/src/main/java/org/teavm/backend/wasm/runtime/net/impl/SockAddrInet6.java
mandesero: I’ve removed these methods for parsing string representations of addresses in `SockAddrInet4/6` since they are not used.

core/src/main/java/org/teavm/backend/wasm/runtime/net/impl/SockAddrInet6.java
mandesero: This code has been removed, so it is no longer relevant.

core/src/main/java/org/teavm/backend/wasm/runtime/net/impl/AddrInfo.java
```diff
@@ -0,0 +1,125 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.backend.wasm.runtime.net.impl;
+
+import org.teavm.backend.wasm.runtime.WasiBuffer;
+import org.teavm.interop.Address;
+import org.teavm.interop.Structure;
+
+public class AddrInfo {
```
konsoletyper, line 22: I still don't understand the purpose of this class. Why don't you just expose `AddrInfoStruct` directly, without any wrappers?

mandesero: I reviewed it again and realized that we can completely remove `AddrInfo`, which would simplify the code. I also simplified the `AddrInfoHints` class into a struct and moved it directly into `WasiVirtualSocket` since it's not used anywhere else.

PR #1014, opened 2025-03-30
Add PropertyChangeSupport with an minimal implemention. 

pizzadox9999:
Hi, I would like to contribute the PropertyChangeSupport class. It is a minimalist implementation. So it has no Serializable support. For the implementation itself I took the public API and implemented it myself. Sometimes as a help I looked at the Apache Harmony Project. Do I need to add the Apache license to every file? Also I hope those classes are far enough from the core.
Best regards pizza_dox_9999


konsoletyper:
> Sometimes as a help I looked at the Apache Harmony Project

Harmony is safe, you can just borrow the code.

> Do I need to add the Apache license to every file

Yes

Also, please run checkstyle locally to see which code style violations you have

PR #1016, opened 2025-04-04
Extracted fixes from old PR

Ihromant:


konsoletyper:
Can you please provide tests?

Ihromant:
Added test. It seems that on wasm-platforms there is intermittent failures that could be related to failures in previous pull request. Also I noticed that wasm-platform tests fail randomly from time to time.

konsoletyper:
What about test for `identityHashCode`?

> It seems that on wasm-platforms there are intermittent failures that could be related to failures in previous pull request

What do you mean?

> Also I noticed that wasm-platform tests fail randomly from time to time.

It's not wasm-platforms, it's all platforms. If you see results carefully, you'll notice that it's due to the browser could not open JS file, which is strange and also it's quite difficult to diagnose. BTW, do you run tests locally?

Ihromant:
> What about test for `identityHashCode`?

Should it be
```java
assertEquals(0, System.identityHashCode(null));
```
? I don't know how to "guess" what identity hashcode is set to ordinary object. I can just check that it is idempotent or so, so there is other test possibility.
```java
Object o = new Object();
int hc = System.identityHashCode(o);
int hc1 = System.identityHashCode(o);
assertEquals(hc, hc1);
```

> What do you mean?

That this are similar errors that you change one class and test starts failing in some unrelated place. This happens in webassembly platforms regularly.

> It's not wasm-platforms, it's all platforms. If you see results carefully, you'll notice that it's due to the browser could not open JS file, which is strange and also it's quite difficult to diagnose. BTW, do you run tests locally?

I'm ignoring wasm platforms because tests are not running locally with
```
Test failed:
./run-wasi.sh: 2: /home/ihromant/.wasmtime/bin/wasmtime: not found

java.lang.RuntimeException: Test failed:
./run-wasi.sh: 2: /home/ihromant/.wasmtime/bin/wasmtime: not found
```
failure.

konsoletyper:
@Ihromant you can ignore WASI only, all other Wasm platforms run in the browser, you don't need special setup. Anyway, there's documentation: https://teavm.org/docs/contributing/testing.html

Ihromant:
Thanks, got it.

Ihromant:
By the way, is there an option to run tests for specific platform only (js, wasm-gc etc.) using gradle property?

konsoletyper:
Yes, there is such an option

Issue #1020, opened 2025-04-24
Substitutor for bootstrap method was not found (used to work in 0.10.2)


ncordon:
For Scala code (version 2.13.16 at least) we used to be able to transpile bytecode using TeaVM `0.10.2`, but when bumping to `0.11.0` we hit the following error repeated a bunch of times:

```
[ERROR] Substitutor for bootstrap method scala.runtime.LambdaDeserialize.bootstrap(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/invoke/MethodHandle;)Ljava/lang/invoke/CallSite; was not found
    at scala.Function1.$deserializeLambda$(-1)
```

I've tracked down the issue to something that was changed in commit https://github.com/konsoletyper/teavm/commit/383fee67c57e5efea3535db711655203b68f04da.

A reproduction can be found at https://github.com/ncordon/teavm-repro/  (transpiles fine with `0.10.2` it breaks with `0.11.0`)

PR #1022, opened 2025-04-25
Replaces lambda serialization in Scala with no-op

ncordon:
Addresses #1020.

## What

It makes `scala.runtime.LambdaDeserialize.bootstrap` behave as no-op. 

My understanding is that `scala.runtime.LambdaDeserialize.bootstrap` would only be used in case we serialize / deserialize a lambda, which is not possible in TeaVM anyway (because there's no implementation for `ObjectInputStream` or `ObjectOutputStream`).

## Why 

In the previous version of the library (`0.10.x`) the `processInvokeDynamic` lived inside the `DependencyAnalyzer` class and it was moved inside `DependencyClassSource` in commit 383fee6. So this does not change the behaviour prior to that commit, where `processInvokeDynamic` could never get called with `scala.runtime.LambdaDeserialize`. Maybe there's a more elegant way to achieve what I want in this PR.

The code in [`samples/scala`](https://github.com/konsoletyper/teavm/tree/master/samples/scala) would not transpile before and now it does by doing `./gradlew :scala:build` inside the `samples` folder.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/impl/EmptyBootstrapSubstitutor.java
+++ b/classlib/src/main/java/org/teavm/classlib/impl/EmptyBootstrapSubstitutor.java
@@ -1,0 +1,29 @@
+/*
+ *  Copyright 2025 Nacho Cordon.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


konsoletyper:
I think there's another solution for that. `DependencyClassHolder` currently processes invoke dynamics in all methods and reports errors immediately. I think it should report them lazily, and then, only when `DependencyAnalyzer` reaches particular method, it should report corresponding errors. At least, it's how now "missing class/method/field" diagnostic works.

ncordon:
I see you addressed the problem in 0347e92, thanks a lot @konsoletyper ! 😄

PR #1058, opened 2025-07-17
Enhance JUnit test run visibility for method/platform/configuration

vellotis:
As I was trying to figure out why WASI and C platform tests failed on my Mac, I felt a crying need to visualize which method+platform+configuration exactly failed. I changed the code for myself just enough so it would make my life easier. Decided to turn this into a PR presentable enhancement.

image

> C platform tests are disabled in the screenshot test run. 

Not sure if you plan to have another `0.12.x` version or this PR should only be introduced into `master`.

### Changes
- Refactor TeaVMTestRunner to maintain test descriptions and test results visibility across multiple platforms and configurations.

  New suite/test descriptors structure:
  ```
  TestClass(Suite)
  ├─  "Compile Platform+Configuration"(Tests)
  └─  Method(Suites)
      └─  Platform+Configuration(Tests)
  ```

- Introduced `JVM` as `org.jvm.junit.TestPlatform` value. It makes creating test `Descriptiors` more unified.
- Introduce visibility for compilation per platform/configuration
  - remove redundant `failAllClasses`. No point for running/failing/ignoring all other tests. Just skip. Compilation failure is already enough.
  - skip (even ignoring) tests that are related to TestClass, if whole class compilation fails
- If JVM test run is not skipped, but fails, all test method related tests are ignored.
- Changed `generateName` method to use `[]` instead of `()`, so it could be used also for test `Description`. `()` are already preserved for other means in test Descriptor display name. Change affects `tests.json` test name format.

### DIFF
--- a/tools/junit/src/main/java/org/teavm/junit/TeaVMTestRunner.java
+++ b/tools/junit/src/main/java/org/teavm/junit/TeaVMTestRunner.java
@@ -33,9 +33,9 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


konsoletyper:
Does this work properly in IDEA?

konsoletyper:
I also mean: if I click "run" next to a certain test method or class, will this approach work? If I do this in test result? AFAIK, usually such custom things aren't supported by IDEs well

PR #1085, opened 2025-10-04
Fixed issue in escaping percent not working.

AsheraCordova:


konsoletyper:
test?

AsheraCordova:
Is there an existing test which I can update.

tests/src/test/java/org/teavm/classlib/java/lang/StringTest.java
```diff
@@ -410,4 +413,23 @@ public void stripIndent() {
         assertEquals("", " ".stripIndent());
         assertEquals("\n", " \n ".stripIndent());
     }
+    
+    @Test
+    public void formatsEscapePercent() {
```
konsoletyper, line 418: Is this duplicate really necessary?

AsheraCordova:
Was not sure if the String.format used Formatter class internally. Hence tested it

PR #1101, opened 2025-10-15
fix(classlib): TClass.java.getFields

tran4o:
use getDeclaredFields() for iteration instead of direct member access (declaredFields). 
Fixes null pointer exception

### DIFF
--- a/tests/src/test/java/org/teavm/classlib/java/lang/reflect/FieldTest.java
+++ b/tests/src/test/java/org/teavm/classlib/java/lang/reflect/FieldTest.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertTrue;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Arrays;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.teavm.classlib.support.Reflectable;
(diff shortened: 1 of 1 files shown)


konsoletyper:
tests?

tran4o:
clear &./gradlew :tests:test --tests org.teavm.classlib.java.lang.reflect.FieldTest --scan

then in the FieldTest.java i wrote the string results to files, stored also the string equals result (true). It worked as run by java (first pass) then the error happens in the compiled JS test.  After succesefull java test with as written text files i removed the test, so the JS version does not have any extra "debug code" 

test was completed in ubuntu 22, jdk 24, with "./gradlew wrapper --gradle-version 8.14.2"

tests/src/test/java/org/teavm/classlib/java/lang/reflect/FieldTest.java
```diff
@@ -142,6 +143,24 @@ public void unusedReflectableField() throws IllegalAccessException {
...
+        Field[] fields = FieldInheritanceSub.class.getFields();
+        assertTrue(fields != null && fields.length == 2);
+
+        Arrays.sort(fields, (a, b) -> a.toString().compareTo(b.toString()));        
+        for (Field field : fields) 
+            sb.append(field.toString()).append(";");
+
+        String shouldBe = "public int org.teavm.classlib.java.lang.reflect.FieldTest$FieldInheritanceBase.base;"
+                + "public java.lang.String org.teavm.classlib.java.lang.reflect.FieldTest$FieldInheritanceSub.sub;";
+        assertTrue(shouldBe.equals(sb.toString()));
```
konsoletyper: assertEquals

konsoletyper:
Also, read this carefully. It can give clues about debugging tests: https://teavm.org/docs/contributing/testing.html

tests/src/test/java/org/teavm/classlib/java/lang/reflect/FieldTest.java
konsoletyper: Just run generated test in the browser. You can also put breakpoints there

PR #1133, opened 2026-01-02
C: Fix windows incorrect SetFilePointer parameter 

xpenatan:
* Fix the where parameter where it does not exist in teavm_file_truncate when its WINDOWS.

### DIFF
--- a/tests/build.gradle.kts
+++ b/tests/build.gradle.kts
@@ -88,7 +88,7 @@ tasks.test {
 
     systemProperty("teavm.junit.c", providers.gradleProperty("teavm.tests.c").orElse("true").get())
     systemProperty("teavm.junit.c.compiler", providers.gradleProperty("teavm.tests.c.compiler")
-        .orElse("compile-c-unix-fast.sh").get())
+            .orElse("compile-c-unix-fast.sh").get())
 
     val dependencies = configurations.testRuntimeClasspath.get()
             .incoming.resolutionResult.allDependencies


tests/build.gradle.kts
```diff
@@ -87,8 +87,12 @@ tasks.test {
             .orElse("./run-wasi.sh").get())
 
     systemProperty("teavm.junit.c", providers.gradleProperty("teavm.tests.c").orElse("true").get())
+    val windowsCompiler = providers.gradleProperty("teavm.tests.c.compiler.windows").orElse("mingw").get()
+    val defaultCCompiler = if (System.getProperty("os.name").startsWith("Windows")) {
```
konsoletyper: The idea behind `teavm.tests.c.compiler` was that user can create ANY actual shell file they want and just put reference to it there. Why not using the same approach for Windows? I.e. set actual shell/bat file based on OS and not introduce any extra properties. Also, I believe VS should be default choice for Windows

core/src/main/resources/org/teavm/backend/c/definitions.h
```diff
@@ -24,9 +24,16 @@
 #endif
 
 #ifdef __GNUC__
-    #undef TEAVM_UNIX
-    #define TEAVM_UNIX 1
-    #include <stdalign.h>
+    #ifdef __MINGW32__
```
konsoletyper: BTW, what the point in using mingw, when there's VS compiler that fits better for developing under windows?

core/src/main/resources/org/teavm/backend/c/definitions.h
```diff
@@ -24,9 +24,16 @@
...
+        #undef TEAVM_UNIX
```
konsoletyper: I believe mingw is more UNIX than Windows. The target is Windows, but available APIs are all POSIX/libc rather than Windows API (which you can use from Windows, but for what reason?).

xpenatan: I removed mingw support so I reverted this change.

PR #1138, opened 2026-01-30
c: Add PSP support

xpenatan:
This pull request introduces PSP code to enable compilation (assisted by AI). The implementation may not be fully correct at this stage, but it can be iteratively improved.

### DIFF
--- a/core/src/main/resources/org/teavm/backend/c/memory.c
+++ b/core/src/main/resources/org/teavm/backend/c/memory.c
@@ -253,7 +253,6 @@ void teavm_initStaticGcRoots() {
 }
 
 #if TEAVM_PSP
-#include <stdlib.h>
 
 int64_t teavm_pageSize() {
     return 4096LL;


core/src/main/resources/org/teavm/backend/c/uchar.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <wchar.h>
```
konsoletyper: I'm not sure this is necessary. At least, not for non-PSP platforms

core/src/main/resources/org/teavm/backend/c/memory.c
```diff
@@ -244,3 +251,22 @@ void teavm_initStaticGcRoots() {
         builder = next;
     }
 }
+
+#if TEAVM_PSP
+
+int64_t teavm_pageSize() {
```
konsoletyper: Don't forget to ping me to improve TeaVM to support moveable heaps (i.e. based on malloc/realloc). Since TeaVM's GC already supports updating references, the same mechanism can be utilized to relocate the heap on platforms that don't support virtual memory

Issue #1156, opened 2026-02-22
TeaVM incorrectly inlines isInfinite as !isFinite


lll69:
In Java, both `isInfinite(Double.NaN)` and `isFinite(Double.NaN)` should return `false`, but TeaVM incorrectly inlines isInfinite as !isFinite, causing `isInfinite(Double.NaN)` to return `true`.

Test Code:
```java
public class InfinityTest {
    public static void main(String[] args) {
        System.out.println("Is NaN Finite: " + Double.isFinite(Double.NaN));
        System.out.println("Is NaN Infinite: " + Double.isInfinite(Double.NaN));
    }
}
```

Run with TeaVM:
```
Is NaN Finite: false
Is NaN Infinite: true
```

Run with Java:
```
Is NaN Finite: false
Is NaN Infinite: false
```

Incorrect inlining can be found in both `org.teavm.classlib.java.lang.TDouble` and `org.teavm.classlib.java.lang.TFloat`.

Issue #1154, opened 2026-02-22
Double.parseDouble and Float.parseFloat cannot parse Infinity, -Infinity, and NaN values


lll69:
It seems that TeaVM's Double.parseDouble and Float.parseFloat cannot parse Infinity, -Infinity, and NaN values.

The following is the test code:
```java
public class ParseDoubleTest {
    private static void tryParseDouble(String str) {
        try {
            double doubleValue = Double.parseDouble(str);
            System.out.println("parseDouble " + str + " Success: " + doubleValue);
        } catch (NumberFormatException e) {
            System.out.println("parseDouble " + str + " Error: " + e.getClass().getName());
        }
    }
    private static void tryParseFloat(String str) {
        try {
            float floatValue = Float.parseFloat(str);
            System.out.println("parseFloat " + str + " Success: " + floatValue);
        } catch (NumberFormatException e) {
            System.out.println("parseFloat " + str + " Error: " + e.getClass().getName());
        }
    }
    private static void tryParse(String str) {
        tryParseDouble(str);
        tryParseFloat(str);
    }
    public static void main(String[] args) {
        System.out.println("--- Constants ---");
        System.out.println(Double.POSITIVE_INFINITY + " " + Float.POSITIVE_INFINITY);
        System.out.println(Double.NEGATIVE_INFINITY + " " + Float.NEGATIVE_INFINITY);
        System.out.println(Double.NaN + " " + Float.NaN);
        System.out.println();
        System.out.println("--- Parse Numbers ---");
        tryParse("Infinity");
        tryParse("+Infinity");
        tryParse("-Infinity");
        tryParse("NaN");
        tryParse("+NaN");
        tryParse("-NaN");
        System.out.println();
        System.out.println("--- Invalid Numbers ---");
        tryParse("INFINITY");
        tryParse("infinity");
        tryParse("++Infinity");
        tryParse("--Infinity");
        tryParse("NAN");
        tryParse("nan");
        tryParse("++NaN");
        tryParse("--NaN");
    }
}
```

Run with TeaVM:
```
--- Constants ---
Infinity Infinity
-Infinity -Infinity
NaN NaN

--- Parse Numbers ---
parseDouble Infinity Error: java.lang.NumberFormatException
parseFloat Infinity Error: java.lang.NumberFormatException
parseDouble +Infinity Error: java.lang.NumberFormatException
parseFloat +Infinity Error: java.lang.NumberFormatException
[...]

PR #1160, opened 2026-02-24
Make Infinity and NaN parsable, make isInfinite(NaN) return false

lll69:
Make Infinity and NaN parsable, and make isInfinite(NaN) return false in JS, fixes issue #1156 on JS and issue #1154


konsoletyper:
Please, fix checkstyle errors

PR #1179, opened 2026-04-29
.

GalacticClient:


konsoletyper:
`Proxy` is not going to work. The functionality which was implemented does not solve the *main* responsibility of this class.

PR #1182, opened 2026-04-29
Fix UnicodeData range parsing for character types

### LINKED ISSUE #1175: `Character.getType` / `\p{L}` does not recognize CJK ideographs inside `<..., First>` / `<..., Last>` ranges from UnicodeData.txt
I found what looks like a Unicode data generation bug in TeaVM classlib.

## Summary

`Character.getType()` and regex Unicode category matching (`\p{L}`) appear to fail for many CJK ideographs, while working for Hiragana, Katakana, and sometimes the first code point of a CJK range.

This breaks code that relies on Unicode letter classification in TeaVM-generated JavaScript.

## Minimal symptom

Examples:

- `Character.getType('あ')` behaves like a letter
- `Character.getType('ア')` behaves like a letter
- `Character.getType('一')` behaves like a letter
- `Character.getType('発')` does **not** behave like a letter
- `Character.getType('漢')` does **not** behave like a letter

Similarly, regex with `\p{L}` matches some Japanese characters but not many Kanji characters.

## Expected behavior

Kanji such as `発` and `漢` should be classified as letters (`OTHER_LETTER`) and should match `\p{L}`.

## Actual behavior

Many Kanji are treated as non-letters / unassigned in TeaVM-generated JavaScript.

## Reproduction idea

Something along these lines should pass:

```java
assertEquals(Character.OTHER_LETTER, Character.getType('あ'));
assertEquals(Character.OTHER_LETTER, Character.getType('ア'));
assertEquals(Character.OTHER_LETTER, Character.getType('一'));
assertEquals(Character.OTHER_LETTER, Character.getType('発'));
assertEquals(Character.OTHER_LETTER, Character.getType('漢'));

assertTrue(Pattern.compile("\\p{L}+").matcher("発").matches());
assertTrue(Pattern.compile("\\p{L}+").matcher("漢").matches());
```

But in TeaVM-generated JS, the checks for `発` / `漢` fail.

## Suspected root cause

The problem seems to be in Unicode data generation, specifically here:

- `org.teavm.classlib.impl.unicode.UnicodeSupport`
- method: `parseUnicodeData(...)`

From reading the implementation, it looks like `UnicodeData.txt` is processed line by line as individual code points, but ranges expressed as:

- `<..., First>`
- `<..., Last>`

are not expanded.

That is important for CJK ideographs, because `UnicodeData.txt` contains entries like:

```text
4E00;<CJK Ideograph, First>;Lo;...
9FFF;<CJK Ideograph, Last>;Lo;...
```

So:

- `U+4E00` (`一`) may work because it is explicitly present as the `First` entry
- interior code points such as `U+767A` (`発`) and `U+6F22` (`漢`) are not explicitly listed
- if the range is not expanded, they end up missing from the generated category tables
- then `Character.getType()` returns the wrong category, and `\p{L}` fails as well

## Proposed fix

In `UnicodeSupport.parseUnicodeData(...)`, detect `<..., First>` / corresponding `<..., Last>` pairs and expand the full range into the generated character class table.

## Impact

This affects any TeaVM code that depends on Unicode category classification, including regex character classes like `\p{L}` and any parser or identifier logic built on top of them.

If useful, I can also prepare a patch and a regression test.

I observed that this is not "all Japanese characters fail": Hiragana and Katakana work, and even `一` works, but Kanji inside the CJK range such as `発` and `漢` fail. That pattern is consistent with `<First>/<Last>` range entries not being expanded.

nazoking:
Fixes #1175.

## Summary

This updates UnicodeData.txt parsing to expand `<..., First>` / `<..., Last>` code point ranges when generating character type tables. These ranges include CJK ideographs, Hangul syllables, private-use areas, and other blocks that previously did not get their proper `Character.getType` values in TeaVM.

It also fixes Unicode RLE compression/extraction edge cases for long runs, so generated Unicode tables round-trip correctly.

## Testing

- `git diff --check origin/master..HEAD`
- `./gradlew :tests:test --tests org.teavm.classlib.java.lang.CharacterTest --tests org.teavm.classlib.java.util.regex.MatcherTest --tests org.teavm.classlib.impl.unicode.UnicodeHelperTest -Pteavm.tests.js=true -Pteavm.tests.wasm=false -Pteavm.tests.wasi=false -Pteavm.tests.c=false --stacktrace`
- `./gradlew build -x test -x javadoc --stacktrace`

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/impl/unicode/UnicodeHelper.java
+++ b/classlib/src/main/java/org/teavm/classlib/impl/unicode/UnicodeHelper.java
@@ -172,7 +172,7 @@ public final class UnicodeHelper {
             byte b = bytes[i];
             if (i < bytes.length - 1 && b == bytes[i + 1]) {
                 int count = 0;
-                while (count < 16384 && bytes[i + count] == b) {
+                while (count < 16384 && i + count < bytes.length && bytes[i + count] == b) {
                     ++count;
                 }
                 i += count;
@@ -219,13 +219,16 @@ public final class UnicodeHelper {
                 count = 1;
             }
             if (b != 0 || count < 128) {
-                if (index + count >= buffer.length) {
-                    ranges[rangeIndex++] = new Range(codePoint, codePoint + index, Arrays.copyOf(buffer, index));
-                    codePoint += index + count;
-                    index = 0;
-                }
-                while (count-- > 0) {
-                    buffer[index++] = b;
+                while (count > 0) {
+                    if (index == buffer.length) {
+                        ranges[rangeIndex++] = new Range(codePoint, codePoint + index, Arrays.copyOf(buffer, index));
+                        codePoint += index;
+                        index = 0;
+                    }
+                    int chunk = Math.min(count, buffer.length - index);
+                    Arrays.fill(buffer, index, index + chunk, b);
+                    index += chunk;
+                    count -= chunk;
                 }
             } else {
                 if (index > 0) {
@@ -234,6 +237,9 @@ public final class UnicodeHelper {
                 codePoint += index + count;
                 index = 0;
             }
+        }
+        if (index > 0) {
+            ranges[rangeIndex++] = new Range(codePoint, codePoint + index, Arrays.copyOf(buffer, index));
         }
         return Arrays.copyOf(ranges, rangeIndex);
     }
--- a/classlib/src/test/java/org/teavm/classlib/impl/unicode/UnicodeHelperTest.java
+++ b/classlib/src/test/java/org/teavm/classlib/impl/unicode/UnicodeHelperTest.java
@@ -18,10 +18,7 @@ package org.teavm.classlib.impl.unicode;
 import static org.junit.Assert.assertArrayEquals;
 import java.util.Arrays;
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.teavm.junit.TeaVMTestRunner;
 
-@RunWith(TeaVMTestRunner.class)
 public class UnicodeHelperTest {
     @Test
     public void longRunsRoundTrip() {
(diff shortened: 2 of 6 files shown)


classlib/src/test/java/org/teavm/classlib/impl/unicode/UnicodeHelperTest.java
```diff
@@ -0,0 +1,39 @@
...
+package org.teavm.classlib.impl.unicode;
+
+import static org.junit.Assert.assertArrayEquals;
+import java.util.Arrays;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.teavm.junit.TeaVMTestRunner;
+
+@RunWith(TeaVMTestRunner.class)
+public class UnicodeHelperTest {
```
konsoletyper, line 22: I believe this is that rare case when you don't need test in `:tests` module. I guess there's might be already something about unicode tests in classlib. Anyway, if there's none, you should move this one there. And you don't need to run it in the browser

classlib/src/main/java/org/teavm/classlib/impl/unicode/UnicodeHelper.java
```diff
@@ -172,7 +172,7 @@ public static String compressRle(byte[] bytes) {
             byte b = bytes[i];
             if (i < bytes.length - 1 && b == bytes[i + 1]) {
                 int count = 0;
-                while (count < 16384 && bytes[i + count] == b) {
+                while (count < 16384 && i + count < bytes.length && bytes[i + count] == b) {
```
konsoletyper, line 175: I'm not sure that was the intention here and why did you edit this file

nazoking: This change is needed because expanding UnicodeData ranges makes the generated character class table end with a long non-zero run, specifically the Plane 16 private-use range. The old compressRle loop can read past the end when a repeated run reaches the end of the byte array, and extractRle also does not flush trailing buffered data. It can additionally advance codePoint too far when splitting a non-zero run at the internal buffer boundary.

So this is not an unrelated cleanup: without it, the corrected UnicodeData range expansion either fails during metadata generation or produces incorrect decoded ranges for Character.getType.

classlib/src/test/java/org/teavm/classlib/impl/unicode/UnicodeHelperTest.java
nazoking: Done, thanks. I moved UnicodeHelperTest from :tests to :classlib and removed TeaVMTestRunner. This test only verifies the RLE helper round-trip behavior, so it now runs as a regular JVM test instead of through the browser runner.

PR #1195, opened 2026-06-03
Support record-style @JSProperty accessors

electrum:
Allow using `@JSProperty` on record-style accessor methods (i.e. no `get` / `is` prefix) without specifying an explicit property name.

### DIFF
--- a/jso/impl/src/main/java/org/teavm/jso/impl/JSPropertyUtil.java
+++ b/jso/impl/src/main/java/org/teavm/jso/impl/JSPropertyUtil.java
@@ -1,0 +1,66 @@
+/*
+ *  Copyright 2026 Alexey Andreev.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.jso.impl;
+
+import org.teavm.model.MethodReader;
+import org.teavm.model.ValueType;
+
+final class JSPropertyUtil {
+    private JSPropertyUtil() {
+    }
+
+    static boolean isProperGetter(MethodReader method) {
+        return method.parameterCount() == 0 && method.getResultType() != ValueType.VOID;
+    }
+
+    static boolean isProperSetter(MethodReader method, String suggestedName) {
+        return method.parameterCount() == 1 && method.getResultType() == ValueType.VOID
+                && (suggestedName != null || isProperPrefix(method.getName(), "set"));
+    }
+
+    static String getGetterName(MethodReader method) {
+        if (method.getResultType() == ValueType.BOOLEAN && isProperPrefix(method.getName(), "is")) {
+            return cutPrefix(method.getName(), 2);
+        }
+        if (isProperPrefix(method.getName(), "get")) {
+            return cutPrefix(method.getName(), 3);
+        }
+        return method.getName();
+    }
+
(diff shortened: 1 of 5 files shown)


jso/impl/src/main/java/org/teavm/jso/impl/JSPropertyUtil.java
```diff
@@ -0,0 +1,83 @@
+/*
+ *  Copyright 2026 Alexey Andreev.
```
konsoletyper, line 2: Strange that you are putting my name here

jso/impl/src/main/java/org/teavm/jso/impl/JSPropertyUtil.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private static String computeGetterName(MethodReader method) {
+        if (method.getResultType() == ValueType.BOOLEAN && isProperPrefix(method.getName(), "is")) {
+            return cutPrefix(method.getName(), 2);
+        }
+        if (isProperPrefix(method.getName(), "get")) {
+            return cutPrefix(method.getName(), 3);
+        }
+        return method.getResultType() != ValueType.VOID ? method.getName() : null;
```
konsoletyper: I believe, this check is not necessary here. You can move it to `isProperGetter` and remove extra null check from `getGetterName`

jso/impl/src/main/java/org/teavm/jso/impl/JSPropertyUtil.java
```diff
@@ -0,0 +1,83 @@
...
+package org.teavm.jso.impl;
+
+import org.teavm.model.MethodReader;
+import org.teavm.model.ValueType;
+
+final class JSPropertyUtil {
+    private JSPropertyUtil() {
+    }
+
+    static boolean isProperGetter(MethodReader method, String suggestedName) {
```
konsoletyper: I believe, `suggestedName` here makes no sence

konsoletyper: Ah, it was in original implementation

jso/impl/src/main/java/org/teavm/jso/impl/JSPropertyUtil.java
electrum: This is mostly moved code. I can put whatever you like in the header.

jso/impl/src/main/java/org/teavm/jso/impl/JSPropertyUtil.java
electrum: Good catch. I simplified the logic and now `suggestedName` is only needed for `isProperSetter`.

jso/impl/src/main/java/org/teavm/jso/impl/JSPropertyUtil.java
electrum: You are correct, `suggestedName` is no longer needed, as any name is legal for record-style getters.

electrum:
This change unifies the implicit property naming logic between `JSObjectClassTransformer` and `JSClassProcessor`, though it introduces backwards incompatibility for exported Java APIs in a few edge cases. The new behavior seems better and more understandable, but I can change it if this is a problem.

## Exported acronym getters

```java
@JSExport @JSProperty
public String getURL() { ... }
```
```javascript
obj.uRL // old name
obj.URL // new name
```

## Exported getter/setter names with non-letters

```java
@JSExport @JSProperty
public String get$foo() { ... }

@JSExport @JSProperty
public void set$foo(String value) { ... }
```
```javascript
// old names
obj.get$foo
obj.set$foo = value

// new names
obj.$foo
obj.$foo = value
```

## Exported invalid-setter methods are no longer prefix-cut

```java
@JSExport @JSProperty
public void setName(String a, String b) { ... }
```
```javascript
obj.name = value    // old name
obj.setName = value // new name
```

konsoletyper:
@electrum when you change something in a PR which I'm already reviewing and I already have commented, *please* add any changes only with new commits. I'll merge them into one commit myself. Otherwise it's not that trivial for me to understand what was changed compared to the code that I already reviewed.

electrum:
@konsoletyper Understood. I'll remember that for future PRs. Thanks for reviewing, and my apologies for making this one harder to review.

PR #1202, opened 2026-07-02
Add System.Logger and friends to class library.

colorizenl:
This was introduced in Java 9 to avoid a hard dependency from the JDK on `java.util.logging` for its own logging. It's comparable to a built-in SLF4J. I think it's rarely used outside of the JDK, but some libraries (in my case SnakeYAML) rely on it.

The `System.Logger` has two possible implementations: If the `java.util.logging` module is available, it uses that. If not, it uses a simple console logger as a fallback. I don't know if I can assume that `java.util.logging` is *always* available in TeaVM, so I'm using the simple console route.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TSystem.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TSystem.java
@@ -18,6 +18,8 @@ package org.teavm.classlib.java.lang;
 import java.lang.reflect.Array;
 import java.util.Enumeration;
 import java.util.Properties;
+import java.util.ResourceBundle;
+import java.util.function.Supplier;
 import org.teavm.backend.c.intrinsic.RuntimeInclude;
 import org.teavm.backend.c.runtime.Memory;
 import org.teavm.backend.c.runtime.fs.CFileSystem;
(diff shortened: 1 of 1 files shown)


classlib/src/main/java/org/teavm/classlib/java/lang/TSystem.java
```diff
@@ -303,4 +305,122 @@ public static String lineSeparator() {
...
+            return ConsoleLogger.INSTANCE;
+        }
+    }
+
+    private static class ConsoleLogger extends LoggerFinder implements Logger {
+
+        private static final ConsoleLogger INSTANCE = new ConsoleLogger();
+
+        @Override
+        public Logger getLogger(String name, Module module) {
```
konsoletyper: Module is not implemented in TeaVM

colorizenl: Argh, didn't realize that. In that case I removed `LoggerFinder` but left `System.Logger`, since that doesn't depend on `Module`. That won't solve every possible library, but it will still work for libraries that only need the logger part, like [the aforementioned SnakeYAML](https://github.com/snakeyaml/snakeyaml/blob/master/src/main/java9/org/yaml/snakeyaml/internal/Logger.java).

PR #1213, opened 2026-07-17
Fix DAY_OF_WEEK when WEEK_OF_MONTH is set (#371)

### LINKED ISSUE #371: DAY_OF_WEEK off by one
`Calendar.set(DAY_OF_WEEK, x)` seems to be result in dates being off by one.

Sample code:

~~~~
Calendar cal = Calendar.getInstance();
cal.setTime(new Date(1544619708631l));
String sDate = "";
for (int x=0;x<8;x++) {
    cal.set(Calendar.DAY_OF_WEEK, x);
    sDate += x + " - " + cal.getTime() + "\n";
}


System.out.println(sDate);
~~~~

Running in Hotspot I get:

~~~~
0 - Sat Dec 15 13:01:48 UTC 2018
1 - Sun Dec 09 13:01:48 UTC 2018
2 - Mon Dec 10 13:01:48 UTC 2018
3 - Tue Dec 11 13:01:48 UTC 2018
4 - Wed Dec 12 13:01:48 UTC 2018
5 - Thu Dec 13 13:01:48 UTC 2018
6 - Fri Dec 14 13:01:48 UTC 2018
7 - Sat Dec 15 13:01:48 UTC 2018
~~~~

Running in TeaVM I get:

~~~~
0 - Fri Dec 14 2018 05:01:48 GMT-0800 (Pacific Standard Time)
1 - Sat Dec 15 2018 05:01:48 GMT-0800 (Pacific Standard Time)
2 - Sun Dec 09 2018 05:01:48 GMT-0800 (Pacific Standard Time)
3 - Mon Dec 10 2018 05:01:48 GMT-0800 (Pacific Standard Time)
4 - Tue Dec 11 2018 05:01:48 GMT-0800 (Pacific Standard Time)
5 - Wed Dec 12 2018 05:01:48 GMT-0800 (Pacific Standard Time)
6 - Thu Dec 13 2018 05:01:48 GMT-0800 (Pacific Standard Time)
7 - Fri Dec 14 2018 05:01:48 GMT-0800 (Pacific Standard Time)
~~~~

Looking at the code and docs, DAY_OF_WEEK accepts constants SUNDAY, ..., SATURDAY with SUNDAY=1 and SATURDAY=7.   We see here that with an input value of `1`, it produces a Saturday, but should produce a Sunday.

egnaro9:
Fixes #371.

`TGregorianCalendar` line 528 uses `days - 2` where every neighbouring branch uses `days - 3` — including `skew`, computed on the line directly above it:

```java
int skew = mod7(days - 3 - (getFirstDayOfWeek() - 1));
days += (fields[WEEK_OF_MONTH] - 1) * 7 + mod7(skew + dayOfWeek - (days - 2)) - skew;
```

Apache Harmony's `GregorianCalendar`, which this is ported from, uses `days - 3` here — the original spans two lines, and the `- 2` looks like a slip when the expression was joined:

```java
days += (fields[WEEK_OF_MONTH] - 1) * 7
        + mod7(skew + dayOfWeek - (days - 3)) - skew;
```

### Effect

With `YEAR=2002, WEEK_OF_MONTH=2` (`America/New_York`, `en_US`): the JVM returns `Sun Jan 06 2002`, TeaVM returns `Sat Jan 12 2002`.

### Tests

`CalendarTest.test_setII` has four assertions that reach this line, commented out since 2015. This re-enables them rather than adding new ones — they fail on master and pass with this change. No live test reached line 528, so the existing suite could not have caught this.

Verified with `-Pteavm.tests.js=true -Pteavm.tests.browser=browser-chrome`; the full Calendar/Date suite stays green.

### Not addressed here

`minimalDaysInFirstWeek` is still unhandled in this branch (no mirror of the `:557-559` logic), so locales with `minDays > 1` remain wrong for `WEEK_OF_MONTH`. Happy to follow up separately — keeping this to the one-line fix.

Found while auditing TeaVM's JVM-vs-browser behaviour: I use TeaVM to ship a Java game engine to the browser, and I look for places where the port and the JVM quietly disagree.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/TGregorianCalendar.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/TGregorianCalendar.java
@@ -525,7 +525,7 @@ public class TGregorianCalendar extends TCalendar {
                 }
                 if (isSet[WEEK_OF_MONTH] && lastDateFieldSet != DAY_OF_WEEK_IN_MONTH) {
                     int skew = mod7(days - 3 - (getFirstDayOfWeek() - 1));
-                    days += (fields[WEEK_OF_MONTH] - 1) * 7 + mod7(skew + dayOfWeek - (days - 2)) - skew;
+                    days += (fields[WEEK_OF_MONTH] - 1) * 7 + mod7(skew + dayOfWeek - (days - 3)) - skew;
                 } else if (isSet[DAY_OF_WEEK_IN_MONTH]) {
                     if (fields[DAY_OF_WEEK_IN_MONTH] >= 0) {
                         days += mod7(dayOfWeek - (days - 3)) + (fields[DAY_OF_WEEK_IN_MONTH] - 1) * 7;
--- a/tests/src/test/java/org/teavm/classlib/java/util/CalendarTest.java
+++ b/tests/src/test/java/org/teavm/classlib/java/util/CalendarTest.java
@@ -106,8 +106,8 @@ public class CalendarTest {
         cal.clear();
         cal.set(Calendar.YEAR, 2002);
         cal.set(Calendar.WEEK_OF_MONTH, 2);
-        //assertTrue("Incorrect result 0d: " + cal.getTime(), cal.getTime()
-        //        .getTime() == 1010293200000L);
+        assertTrue("Incorrect result 0d: " + cal.getTime(), cal.getTime()
+                .getTime() == 1010293200000L);
 
         cal.clear();
         cal.set(Calendar.YEAR, 2002);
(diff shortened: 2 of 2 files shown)


egnaro9:
Hi Alexey — Erik, the author of this PR.

Unusual ask, and a no is completely fine. I'm applying to the Anthropic Fellows Program (AI Security stream), which requires three references who can speak to technical work. Would you be willing to be listed as one?

I'd describe our interaction on the form exactly as it was: a single merged pull request, not a sustained collaboration. If they contact you it would only be about this contribution. They say they may reach out without advance notice and ask for a reply within a week.

If you'd rather not, no reply needed at all — I won't ask again. I can also be reached at erik.hill0987@gmail.com if a private channel is easier.

Either way, thank you for maintaining TeaVM. Finding an assertion that had been disabled since 2015 and getting a one-character fix merged was a genuinely good afternoon.

PR #1224, opened 2026-08-08
Wasm GC: stale stack depth in the coroutine transformation

amritk:
WasmTypeInference updated depthBeforeLastInstructionOut only in instructions that
push a result. After an instruction that pushes nothing the field kept a depth
recorded by some earlier instruction, which may exceed the current stack depth.

CoroutineTransformation reads the field right after visiting a suspending
instruction and uses it to index the stack snapshot taken just before it. When
the suspending instruction is an untyped loop that follows a branch or a throw,
the field still holds the depth from before the stack was emptied, and
createSaveInstructions throws IndexOutOfBoundsException. Kotlin coroutine state
machines produce this shape. 

Now every visitor leaves the field equal to the depth of the stack it produced.

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/model/instruction/WasmTypeInference.java
+++ b/core/src/main/java/org/teavm/backend/wasm/model/instruction/WasmTypeInference.java
@@ -30,6 +30,7 @@ public class WasmTypeInference implements WasmInstructionVisitor {
     @Override
     public void visit(WasmUnreachable instruction) {
         typeStack.clear();
+        depthBeforeLastInstructionOut = 0;
     }
 
     @Override
@@ -39,6 +40,8 @@ public class WasmTypeInference implements WasmInstructionVisitor {
             popN(type.getInputTypes().size());
             depthBeforeLastInstructionOut = typeStack.size();
             typeStack.addAll(type.getOutputTypes());
+        } else {
+            depthBeforeLastInstructionOut = typeStack.size();
         }
     }
 
@@ -50,12 +53,15 @@ public class WasmTypeInference implements WasmInstructionVisitor {
             popN(type.getInputTypes().size());
             depthBeforeLastInstructionOut = typeStack.size();
             typeStack.addAll(type.getOutputTypes());
+        } else {
+            depthBeforeLastInstructionOut = typeStack.size();
         }
     }
 
     @Override
     public void visit(WasmBranch instruction) {
         pop();
+        depthBeforeLastInstructionOut = typeStack.size();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/teavm/backend/wasm/model/instruction/WasmTypeInference.java
```diff
@@ -23,13 +23,19 @@ public class WasmTypeInference implements WasmInstructionVisitor {
     public final List<WasmType> typeStack = new ArrayList<>();
     private int depthBeforeLastInstructionOut;
 
+    /**
```
konsoletyper: Please, remove this comment

core/src/main/java/org/teavm/backend/wasm/model/instruction/WasmTypeInference.java
```diff
@@ -70,6 +81,8 @@ public void visit(WasmNullBranch instruction) {
             if (ref.isNullable()) {
                 typeStack.set(typeStack.size() - 1, ref.asNonNull());
             }
+        } else {
```
konsoletyper: This is excessive. The only case when we get here is that we already had a broken module. Assigning 0 to `depthBeforeLastInstructionOut` makes no sense, becase we can't break already broken module

core/src/main/java/org/teavm/backend/wasm/transformation/CoroutineTransformation.java
```diff
@@ -80,6 +80,24 @@ public CoroutineTransformation(WasmFunctionTypes functionTypes, BaseWasmFunction
     }
 
     public void transform(WasmFunction function) {
+        // The instance is reused for every suspending method, so its per-function state has to be
```
konsoletyper: If your LLM's idea was to just guard cleanup with `finally`, it could have wrapped it in the existing method. Otherwise, it looks strange that the setup, that we perform in one function (`transformImpl`) is cleaned up in another one (`transform`). But this change is not relevant in this particular commit, since we don't deal with an exception, thrown from `CoroutineTransformation`, but rather with invalid module.

tests/src/test/java/org/teavm/vm/WasmAsyncTestGenerator.java
```diff
@@ -35,11 +35,23 @@ public WasmAsyncTestGenerator(WasmGCCodeGenContext context) {
...
-            var param = new WasmLocal(WasmType.INT32, "n");
-            function.add(param);
-            var generator = new Generator(context, param);
-            generator.generate(function.getBody().builder());
+        var param = new WasmLocal(WasmType.INT32, "n");
+        var generator = new Generator(context, param);
+        switch (method.getName()) {
+            case "generatedMethod":
+                function.add(param);
+                generator.generate(function.getBody().builder());
```
konsoletyper: It's strange that you added several generators to the `Generator` class, but did not rename original one. Also, I doubt that you really need to put generation logic for `loopAfterBranch` into `Generator` class, since it's not using any of its routines.

amritk:
@konsoletyper is this better? sorry been a long time since I did any java

PR #1225, opened 2026-08-08
Wasm GC: local.tee pushes the local's type, not the value's

amritk:
The spec defines local.tee as local.set followed by local.get, so what stays on
the stack is the local's declared type. WasmTypeInference left the type of the
value instead. The two differ whenever a value is stored into a local declared at
a supertype, which register allocation produces routinely.

CoroutineTransformation reads that stack to build the block signature its resume
path branches to. A subclass type recorded for a superclass value gives a label
typed at the subclass with a superclass value branching to it: "type error in
branch[0] (expected (ref null 342), got (ref null 40))". Compiling the IntelliJ
platform reaches this in java.lang.System.getProperty, where java.util.Properties
is recorded for a java.util.Hashtable local.

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/model/instruction/WasmTypeInference.java
+++ b/core/src/main/java/org/teavm/backend/wasm/model/instruction/WasmTypeInference.java
@@ -150,7 +150,9 @@ public class WasmTypeInference implements WasmInstructionVisitor {
 
     @Override
     public void visit(WasmTeeLocal instruction) {
-        depthBeforeLastInstructionOut = typeStack.size() - 1;
+        pop();
+        depthBeforeLastInstructionOut = typeStack.size();
+        typeStack.add(instruction.getLocal().getType());
     }
 
     @Override
--- a/tests/src/test/java/org/teavm/vm/WasmTeeTypeDependency.java
+++ b/tests/src/test/java/org/teavm/vm/WasmTeeTypeDependency.java
@@ -1,0 +1,35 @@
+/*
+ *  Copyright 2026 Alexey Andreev.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


konsoletyper:
This PR does just the opposite that it states

* Current behaviour: preserve the value's type as is, which conforms to the spec (https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-variable-mathsf-local-tee-x)
* Behaviour after the PR: pop value's type, push local's type, which is not correct.

Apparently, you LLM did not determine the root issue properly and then produced wrong "fix".

konsoletyper:
Ah, sorry. Found more relevant section in the spec: https://webassembly.github.io/spec/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-variable-mathsf-local-tee-x

So that this change is correct, but the description was a bit misleading. "local.tee writes a value into a local and pushes it back, and what it pushes is the local's declared type" - sounds as "how actual code is behaving"

amritk:
> Ah, sorry. Found more relevant section in the spec: https://webassembly.github.io/spec/core/valid/instructions.html#xref-syntax-instructions-syntax-instr-variable-mathsf-local-tee-x
> 
> So that this change is correct, but the description was a bit misleading. "local.tee writes a value into a local and pushes it back, and what it pushes is the local's declared type" - sounds as "how actual code is behaving"

Sorry was in a bit of a rush, will come back and clean these up!

konsoletyper:
Please, review the code after agent. The previous PR fixes issue in same file and introduces new test method in a class that's intended for such tests, but in this PR you announce a new test class without obvious reason. This also includes compiler plugins written for this test.

amritk:
You are right, I will take a look properly when I get some time and probably put up a few more PRs

PR #1226, opened 2026-08-08
Wasm GC: a flattened conditional leaves the type it declares

amritk:
processOptimizedConditional replaces a conditional containing a suspension point
with a block, walking one arm and then the other with the same type inference, so
the recorded stack ends up holding whatever the arm walked last pushed. What the
emitted block actually leaves behind is its declared output type.

The two differ when the arms push different subtypes of the conditional's type:
the block is declared to produce Object, the else arm pushes a String, and String
is what gets recorded. A suspension point further down the same list turns that
stack into the block signature its resume path branches to, so the label ends up
typed at String while an Object reaches it: "type error in branch[1] (expected
(ref null 37), got (ref null 5))".

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/transformation/CoroutineTransformation.java
+++ b/core/src/main/java/org/teavm/backend/wasm/transformation/CoroutineTransformation.java
@@ -329,6 +329,12 @@ public class CoroutineTransformation {
                 replacement.visit(wrapper);
             }
 
+            // Whichever arm was walked last left its own types on the stack, but what the wrapper
+            // block leaves behind is the type it is declared to produce. Recording the arm's types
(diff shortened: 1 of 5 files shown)


konsoletyper:
The same complaint about tests here

core/src/main/java/org/teavm/backend/wasm/transformation/CoroutineTransformation.java
```diff
@@ -329,6 +329,12 @@ private void processOptimizedConditional(WasmConditional conditional) {
                 replacement.visit(wrapper);
             }
 
+            // Whichever arm was walked last left its own types on the stack, but what the wrapper
```
konsoletyper, line 332: Please, avoid such comments. I believe, they quickly become outdated.

PR #1228, opened 2026-08-13
wasm gc: split the module initializer into separate functions

barisyild:
Large applications produce a start function that no engine will compile.

In the app I hit this with, class initialization, string pool initialization and reflection metadata are all appended into the module's start function, which grew to **3.7 MB of code in a single function** — 1,027,444 top-level instructions, 99.99% of the whole initializer. V8 then refuses to instantiate the module:

```
A single code object needs more than half of the code space size
```

which arrives as a SIGTRAP (exit 133) under Node and as a failed instantiation in Chromium. The threshold depends on the platform's code space, so the same module could still load on macOS while dying on Linux, which made it look like a memory problem for a while.

### Change

Each contributor now emits into its own functions and the initializer only calls them, in the same order as before:

* `WasmGCClassGenerator` → `teavm@initClasses@partN`
* `WasmGCStringPool` → `teavm@initStrings@partN`
* `ReflectionMetadataGenerator` → `teavm@initReflection@partN`

The split has to happen while the statements are produced, not by slicing the finished initializer afterwards: within a class, and within a string batch that pushes N values before consuming them with `array.new_fixed`, the statements push and consume values on the operand stack, so cutting at an arbitrary point leaves values behind and the module fails to validate with `expected 0 elements on the stack for fallthru`. A class (and a string batch) is the smallest self-contained unit, so parts are rotated on that boundary once the instruction budget is exceeded.

While here, `initializerFunctionStatements` is removed: nothing writes to it once class initializers go into parts, so the `transferFrom` was always transferring an empty list.

### Result

On the same application the largest function went from 3,768,336 bytes to 93,620 bytes and the module instantiates. Output is otherwise unchanged: same instructions, same order, one call indirection per part.

### Notes for review

* The budgets (4096 / 6144 instructions) only decide how many parts get produced; any value well below the engine limit works.
* Counting the instructions a class contributed means walking the list, because `WasmInstructionList` exposes only `isEmpty()`. If you would rather have a maintained `size` on the list, I am happy to do that instead.
* Small modules now also get one part function each. If you want the initializer to stay as-is below a threshold, that is a small addition.
* I did not add a test — reproducing it needs a module large enough to cross the engine limit. Suggestions welcome.

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/generate/classes/WasmGCClassGenerator.java
+++ b/core/src/main/java/org/teavm/backend/wasm/generate/classes/WasmGCClassGenerator.java
@@ -48,7 +48,6 @@ import org.teavm.backend.wasm.model.WasmStructure;
 import org.teavm.backend.wasm.model.WasmType;
 import org.teavm.backend.wasm.model.instruction.WasmFunctionReference;
 import org.teavm.backend.wasm.model.instruction.WasmInstructionBuilder;
-import org.teavm.backend.wasm.model.instruction.WasmInstructionList;
 import org.teavm.backend.wasm.model.instruction.WasmNullCondition;
 import org.teavm.backend.wasm.model.instruction.WasmNullConstant;
 import org.teavm.backend.wasm.model.instruction.WasmSignedType;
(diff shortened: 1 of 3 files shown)


konsoletyper:
I'm interested if you are using TeaVM with optimizations turned off + some additional pass with external tool. In case you do so, I strongly suggest you NOT TO.

Also, as for this issue. I asked you if you can estimate what is taking the most space in initializer, but you have never answer. I believe, though, it's virtual tables. If it's the case, then applying optimization should decrease them dramatically. Or it does not help?

barisyild:
> I'm interested if you are using TeaVM with optimizations turned off + some additional pass with external tool. In case you do so, I strongly suggest you NOT TO.
> 
> Also, as for this issue. I asked you if you can estimate what is taking the most space in initializer, but you have never answer. I believe, though, it's virtual tables. If it's the case, then applying optimization should decrease them dramatically. Or it does not help?

The optimizations resolved the issue; I am updating the issue right away.

PR #1233, opened 2026-08-14
feat: add webgpu support

arlenhaghighatjou:
Added WebGPU support. It makes adapters, devices, buffers, textures, shaders, pipelines, render/compute passes, queues, canvas contexts, constants, and descriptors.

### DIFF
--- a/jso/apis/src/main/java/org/teavm/jso/webgpu/GPUBindGroup.java
+++ b/jso/apis/src/main/java/org/teavm/jso/webgpu/GPUBindGroup.java
@@ -1,0 +1,8 @@
+package org.teavm.jso.webgpu;
+
+public interface GPUBindGroup extends GPU.ObjectBase {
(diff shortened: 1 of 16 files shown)


jso/apis/src/main/java/org/teavm/jso/webgpu/GPUBindGroup.java
```diff
@@ -0,0 +1,8 @@
+package org.teavm.jso.webgpu;
+
+public interface GPUBindGroup extends GPU.ObjectBase {
+    interface Layout extends GPU.ObjectBase {
+        // le ✨stub✨
```
konsoletyper: I believe such comments are inappropriate

arlenhaghighatjou:
is it good now?

arlenhaghighatjou:
Thank you!!

PR #1235, opened 2026-08-14
model: use the full binary name for inner-class erasure in GenericValueType.asValueType

### LINKED ISSUE #1232: JS: reflection field metadata emits unqualified identifier for inner-class field types (ReferenceError at eval)
### Summary

When a `ReflectionSupplier` grants field access to a class that has a field whose type is an **inner class**, the JS backend's reflection metadata writer emits the field's generic-type descriptor with a bare, unqualified simple name — producing an invalid script that throws `ReferenceError` during evaluation, before `main` runs. Same failure family as #1210 (invalid identifier in emitted reflection metadata), different writer path.

Version: 0.15.0, JS backend, `OptimizationLevel.NONE`, macOS, temurin-17.

### Observed output

Granting fields of `java.lang.ref.ReferenceQueue$RemoveCallback` (its `next` field has type `RemoveCallback`) produces:

```js
["next", 0, jlr_ReferenceQueue$RemoveCallback, o => o.$next11, (o, v) => o.$next11 = v, {t: [6, RemoveCallback]}],
```

The owner slot is correctly `jlr_ReferenceQueue$RemoveCallback`, but the type-descriptor slot (`t:` — kind 6, RAW_TYPE) references `RemoveCallback`, an identifier that does not exist in the script:

```
ReferenceError: RemoveCallback is not defined
    at .../main.js:54197:109
```

### Repro sketch

Any `ReflectionSupplier` returning field names for a class with an inner-class-typed field, e.g. granting `java.lang.ref.ReferenceQueue`'s internals, or in user code:

```java
class Outer { static class Inner {} Inner field; }
```

with `getAccessibleFields` returning `["field"]`.

### Impact

Blanket reflection grants (useful for GSON-style reflective serialization of model packages) are unusable — one inner-typed field anywhere in the granted set invalidates the whole bundle at load time.

thesupersupersigma:
Fixes #1232.

`GenericValueType.Object.asValueType()` built the erasure from the segment-only `className`, so a chained inner type parsed from `Ljava/lang/ref/ReferenceQueue<TT;>.RemoveCallback;` erased to `ValueType.object("RemoveCallback")` — a class that does not exist. Downstream, `ClassReflectionInfoGenerator.generateGenericType`'s raw branch feeds that into `RenderingUtil.typeToClsString` → `appendClass`, which mints a fresh alias for the unknown name and emits reflection field metadata referencing an undefined identifier:

```js
["next", 0, jlr_ReferenceQueue$RemoveCallback, ..., {t: [6, RemoveCallback]}],
```

→ `ReferenceError: RemoveCallback is not defined` at script evaluation, before `main`.

The fix uses `getFullClassName()` (parent-aware binary name); the erasure of `Outer<X>.Inner` is `Outer$Inner`. Unit test covers the chained and the top-level case.

### DIFF
--- a/core/src/main/java/org/teavm/model/GenericValueType.java
+++ b/core/src/main/java/org/teavm/model/GenericValueType.java
@@ -224,7 +224,7 @@ public abstract class GenericValueType {
 
         @Override
         public ValueType asValueType() {
-            return arguments == null || arguments.length == 0 ? ValueType.object(className) : null;
+            return arguments == null || arguments.length == 0 ? ValueType.object(getFullClassName()) : null;
         }
 
         @Override
--- a/core/src/test/java/org/teavm/model/GenericValueTypeTest.java
+++ b/core/src/test/java/org/teavm/model/GenericValueTypeTest.java
@@ -1,0 +1,37 @@
+/*
+ *  Copyright 2026 thesupersupersigma.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


konsoletyper:
Can you please add end-to-end test for this issue?

PR #1236, opened 2026-08-14
js: escape JavaScript reserved words in DefaultAliasProvider

### LINKED ISSUE #1234: JS: class named after a JS reserved word ("in") produces an unparseable bundle
### Summary

A class in the default package whose name is a JavaScript reserved word produces an invalid script on the JS backend: `DefaultAliasProvider.sanitize` checks characters only, so a class named `in` gets the alias `in` and the output contains constructs like:

```js
$$$8 = new in;
```

→ `SyntaxError: Unexpected token 'in'` — the entire bundle fails to parse, in every engine.

Version: 0.15.0, JS backend, `obfuscated=false`.

### How this happens in practice

Obfuscated applications: name obfuscators emit single short bytecode names that are perfectly legal in class files but reserved in JS (`in` and `with` are even legal Java source identifiers). Compiling an obfuscated app (in our case Minecraft 1.20.1, which ships a top-level class named `in`) produces an unparseable bundle. Packaged classes are unaffected because package abbreviation prefixes the alias (`jl_...`).

`MinifyingAliasProvider` appears to have no reserved-word handling either — short generated aliases could presumably collide with `in`/`do`/`if`/etc. the same way (not yet observed; our reproduction is the default provider).

### Suggested fix

Reserved-word check at the end of `sanitize` (applies to every alias kind; member aliases are `$`-prefixed and unaffected). PR follows.

thesupersupersigma:
Fixes #1234.

`DefaultAliasProvider.sanitize` validates characters only, so a default-package class named after a JS reserved word (`in`, `with` — both legal Java identifiers, common in name-obfuscated applications; Minecraft ships a top-level class named `in`) becomes its own alias and the output contains `$$$8 = new in;` — a `SyntaxError` that makes the whole bundle unparseable in every engine.

Reserved results now get a trailing underscore inside `sanitize`, which covers every alias kind that flows through `makeUnique`/`makeUniqueInstance`; instance aliases are `$`-prefixed and unaffected, and the uniqueness counter keeps escaped names collision-free against real classes named `in_`.

`MinifyingAliasProvider` may want the same treatment for generated short names; this PR only fixes the default provider, where we can reproduce.


core/src/main/java/org/teavm/backend/javascript/codegen/DefaultAliasProvider.java
```diff
@@ -153,6 +153,14 @@ private ScopedName makeUnique(String suggested) {
         return new ScopedName(alias, additionalScopeStarted);
     }
 
+    private static final Set<String> RESERVED_WORDS = Set.of(
```
konsoletyper: There's a list of keywords in `org.teavm.backend.javascript.rendering.RenderingUtil`

konsoletyper:
I believe the more efficient strategy would be to increase `knowAliasesCounter` for all keywords. This will automatically turn `in` into `in_1`

thesupersupersigma:
Hey, I genuinely didn't know this PR (and the other two) existed until a few hours ago. Im sorry about that.

Updated based on your feedback, reuses RenderingUtil's keyword list now instead of copying it, and seeds knowAliasesCounter per keyword so makeUnique handles the rest. Turns out that's in1, not in_1, no separator in the concatenation. MinifyingAliasProvider already guards this on its own, left it alone.

One thing though, the counter gets cleared once scope passes maxTopLevelNames, so a keyword class aliased after that point wouldn't be covered. Want me to patch that too, or handle it separately

core/src/test/java/org/teavm/backend/javascript/codegen/DefaultAliasProviderTest.java
```diff
@@ -0,0 +1,36 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.backend.javascript.codegen;
+
+import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+
+public class DefaultAliasProviderTest {
```
konsoletyper: I believe this test is not useful at all. It makes no sense to test that

konsoletyper:
> One thing though, the counter gets cleared once scope passes maxTopLevelNames, so a keyword class aliased after that point wouldn't be covered

Yes, please. It should be trivial

thesupersupersigma:
Pushed. seedKeywords() is called from both the constructor and the scope transition now.

Though looking closer I don't think the gap I flagged is actually reachable. After the transition names are scoped and render as property accesses, where reserved words are legal, so the bare binding form only happens pre-transition and the constructor seeding already covers it. Kept the call anyway so the counter stays valid across the clear(), but happy to drop it and keep the diff at 4 lines if you'd rather.

Issue #1210, opened 2026-07-16
JS: $rt_simpleConstructors references optimizer-eliminated constructors of abstract classes, breaking ADVANCED builds


Ponchia:
### Summary

When a JS build uses `ADVANCED` optimization, the emitted `$rt_simpleConstructors([...])` table can reference constructor functions that the optimizer eliminated, producing an invalid script. In a plain deployment the bundle dies with `ReferenceError: <name> is not defined` during script evaluation, before `main` runs.

### Root cause

`ClassInfoGenerator.writeSimpleConstructors` emits a `(class, <init>)` pair for every class that reaches the `Class.newInstance` dependency node, filtering only on the constructor itself:

```java
if (ctor != null && !ctor.hasModifier(ElementModifier.ABSTRACT)
        && ctor.getLevel() == AccessLevel.PUBLIC) {
```

A constructor is never `ABSTRACT`, and a public abstract class has a public implicit constructor, so **abstract classes pass this filter**. Their no-arg constructor is only ever called through subclass `super()` chains; at `ADVANCED` the inliner absorbs those calls and eliminates the standalone function, while the metadata still references it by name.

This is also inconsistent with the runtime: `TClass.newInstance()` already throws `InstantiationException` for `ABSTRACT`/`INTERFACE` classes *before* consulting the constructor table, so these entries are unreachable dead weight at any optimization level.

### Reproduction

Any abstract class with a public no-arg constructor that (a) flows into `Class.newInstance` and (b) whose constructor is only reachable via subclass `super()` calls, compiled at `ADVANCED`:

```java
public abstract static class AbstractParent {
    public AbstractParent() { }
    abstract String name();
}
static class ConcreteChild extends AbstractParent {
    String name() { return "concrete"; }
}
...
new ConcreteChild().name();
AbstractParent.class.newInstance(); // any newInstance call site reachable from the class
```

In the TeaVM test harness the same defect surfaces as `TypeError: m is not a function` in the optimized/minified configuration (the table holds a dangling reference).

### Real-world impact

[...]

PR #1238, opened 2026-08-14
reflection: link types referenced by accessible member metadata

### LINKED ISSUE #1237: JS: reflection metadata references unlinked member types under ReflectionSupplier grants (ReferenceError at eval)
### Summary

When a `ReflectionSupplier` grants access to a member whose parameter, result or field type is a class nothing else in the program uses, the JS backend emits reflection metadata referencing that class's alias — but the class itself is never linked, so the identifier does not exist in the output and the bundle throws `ReferenceError` during evaluation, before `main`.

Version: 0.15.0, JS backend, `OptimizationLevel.NONE`.

### Observed output

Granting `java.lang.Thread`'s constructors (blanket supplier) produces:

```js
["<init>", 1, $rt_voidcls, [jl_ThreadGroup, jl_Runnable, jl_String], jl_Thread__init_3]
```

with `jl_ThreadGroup` never defined anywhere in the script:

```
ReferenceError: jl_ThreadGroup is not defined
```

Same family as #1210/#1232 — reflection metadata referencing entities absent from the output — but a different gap: `ReflectionDependencyListener` links the granted members themselves, never the types their signatures mention.

### Suggested fix

Link the object base of each granted member's parameter types, result type and field type where accessibility is materialized (`gatherAccessibleFields` / `gatherAccessibleMethods`, which every access path funnels through via the caches). PR follows.

thesupersupersigma:
Fixes #1237.

`ReflectionDependencyListener` links granted members but never the classes their signatures mention, so metadata like

```js
["<init>", 1, $rt_voidcls, [jl_ThreadGroup, jl_Runnable, jl_String], jl_Thread__init_3]
```

references `jl_ThreadGroup` without any definition when nothing else in the program uses `ThreadGroup` — `ReferenceError` at script evaluation. Same family as #1210/#1232: reflection metadata pointing at entities absent from the output.

This links the object base of each accessible member's parameter types, result type and field type inside `gatherAccessibleFields`/`gatherAccessibleMethods` — the cached choke point every enumeration and access path funnels through — so the referenced classes exist in the output even when reflection metadata is their only user.

Reproduced with a blanket ReflectionSupplier over a small app (any granted constructor with a parameter type that is otherwise unused); verified against a large obfuscated app where the blanket grant previously produced an unparseable bundle.

### DIFF
--- a/core/src/main/java/org/teavm/reflection/ReflectionDependencyListener.java
+++ b/core/src/main/java/org/teavm/reflection/ReflectionDependencyListener.java
@@ -359,7 +359,7 @@ public class ReflectionDependencyListener extends AbstractDependencyListener {
                     ClassReader cls = agent.getClassSource().get(className);
                     if (cls != null) {
                         var skipPrivates = shouldSkipPrivates(cls);
-                        var reflectable = collectAccessibleMethods(className);
+                        var reflectable = collectAccessibleMethods(agent, className);
                         for (MethodReader reflectableMethod : cls.getMethods()) {
                             if (!reflectable.contains(reflectableMethod.getDescriptor())) {
                                 continue;
@@ -685,7 +685,7 @@ public class ReflectionDependencyListener extends AbstractDependencyListener {
                 return;
             }
 
-            Set<MethodDescriptor> accessibleMethods = collectAccessibleMethods(className);
+            Set<MethodDescriptor> accessibleMethods = collectAccessibleMethods(agent, className);
 
             var hasConstructors = false;
             for (MethodDescriptor methodDescriptor : accessibleMethods) {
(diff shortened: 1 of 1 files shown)


thesupersupersigma:
One residual worth noting for reviewers: linking only helps for types that exist on the classpath. A granted member can also reference a type with no class definition at all — TeaVM's own `Thread` declares JDK-shaped constructors whose `ThreadGroup` parameter type is absent from the classlib, so blanket grants still emit a dangling `jl_ThreadGroup` reference. The updated branch guards the linker against reader-null types, but the metadata writer itself likely wants the same guard (skip member entries whose types have no reader) so metadata can never reference what cannot exist. Happy to extend this PR or file separately, whichever you prefer.

konsoletyper:
I believe this is a wrong fix. `ReflectionDependencyListener` already declares `addReflectableMethod` right for the same purpose. If it does not work, the reason should be fixed instead of introduction of this extra logic in a wrong place.

konsoletyper:
I also believe that the root reason is the absence of ThreadGroup emulation. So the correct behaviour would be to produce broken module and produce error messages, as it's done in case of non-reflection access to missing class.

PR #1239, opened 2026-08-19
wasm gc: keep the shared reference writer converter type-agnostic

lucas-viva:
Disclaimer: PR created with Claude Fable 5's help.

Reflectively setting fields of two different reference types in the same program fails on the
Wasm GC backend. `ReflectionMetadataGenerator` generates the writer converter for reference
fields once and reuses it for every reference field, but its body casts the incoming value to
the field type of whichever field it was generated for first. Setting any other reference field
then traps:

```
RuntimeError: illegal cast
    at teavm@writerConverter_reference
```

The raw setter generated per field already casts the value to its own field's type, so the
shared converter does not need to narrow at all: this change makes it pass reference values
through as `Object` and only unbox primitives.

The new `FieldTest.setReferenceFieldsOfDifferentTypes` sets a `String` field and then a
`StringBuilder` field through `java.lang.reflect.Field`; without the fix it fails with the trap
above, with the fix it passes. Ran `FieldTest` on the JS and Wasm GC backends and
`checkstyleMain`/`checkstyleTest`.

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/intrinsics/reflection/ReflectionMetadataGenerator.java
+++ b/core/src/main/java/org/teavm/backend/wasm/intrinsics/reflection/ReflectionMetadataGenerator.java
@@ -851,9 +851,13 @@ public class ReflectionMetadataGenerator {
         function.add(boxedValueVar);
 
         var body = function.getBody().builder();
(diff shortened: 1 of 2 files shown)


lucas-viva:
Thank you!

PR #1245, opened 2026-08-25
wasm gc: type inference leaves local.tee at the local's type

lucas-viva:
`WasmTypeInference` keeps the type of the stored value on its stack after a `local.tee`, but
validation types the instruction's result as the local's declared type. When a value of a
narrower type is teed into a wider local (a `String` into an `Object` local) and stays on the
stack across a suspension point, the coroutine transformation snapshots the narrower type into
the signature of the restore block, and the branch the normal path takes across it then carries
the wider type:

```
CompileError: WebAssembly.compileStreaming(): Compiling function
"...::teeThenSuspend" failed: type error in branch[0]
(expected (ref null $java.lang.String), got (ref null $java.lang.Object))
```

This change makes the inference set the stack top to the local's type, matching validation.

The test follows the pattern of `WasmAsyncTest`: a generated body tees a `String` into an
`Object` local, keeps the value on the stack across a suspending call, and consumes it
afterwards. Without the fix the module fails validation with the error above; with it, the test
passes and the value survives suspend/restore. Ran the `org.teavm.vm.*` and `org.teavm.jso.*`
suites on Wasm GC and `checkstyleMain`/`checkstyleTest`.

### DIFF
--- a/tests/src/test/java/org/teavm/vm/WasmTeeLocalTest.java
+++ b/tests/src/test/java/org/teavm/vm/WasmTeeLocalTest.java
@@ -1,0 +1,69 @@
+/*
+ *  Copyright 2026 Alexey Andreev.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


konsoletyper:
BTW, there's another PR that addresses this: [1225](https://github.com/konsoletyper/teavm/pull/1225)

tests/src/test/java/org/teavm/vm/WasmTeeLocalTest.java
```diff
@@ -0,0 +1,69 @@
+/*
```
konsoletyper, line 1: It makes totally no sense to introduce a new test class, while an existing one could be reused

lucas-viva:
Thank you for reviewing it. Closing as a duplicate of the PR you mentioned.

PR #1246, opened 2026-08-25
wasm gc: don't cache reflection call arguments in a non-defaultable local

lucas-viva:
Compiling a program that reflectively invokes a member with more than one parameter, where
that member is suspendable, produces a module that fails WebAssembly validation:

```
CompileError: WebAssembly.compileStreaming(): Compiling function #2994:
"com.badlogic.gdx.scenes.scene2d.ui.SplitPane::<..." failed:
uninitialized non-defaultable local: 2 @+845479
```

The caller functions generated for reflective invocation cache the argument array's data
field in a local when the target member has more than one parameter. That local is
non-defaultable, and when the target method is suspendable the caller goes through the
coroutine transformation, which restores locals inside the resume arm of a conditional — a
non-defaultable local's initialization status does not survive the end of the block that
assigned it.

Regular method bodies already avoid this: `WasmGCMethodGenerator` disables non-nullable
locals for suspendable methods. `generateCaller` builds its wasm body directly and bypasses
that guard. This PR reads the argument array through the (always initialized) `args`
parameter at each use instead of caching it; the cost is one extra `struct.get` per argument
in reflective calls. The coroutine transformation itself is unchanged.

The test is an ordinary reflective call — a `@Reflectable` two-parameter method that sleeps,
invoked via `Method.invoke` — and fails on current master with:

```
CompileError: WebAssembly.compileStreaming(): Compiling function #153:
"org.teavm.vm.WasmAsyncTest$ReflectiveTarget::co..." failed:
uninitialized non-defaultable local: 2
```

End-to-end reproduction with a libGDX game:
https://github.com/lucas-viva/gdx-teavm-1246-repro

### DIFF
--- a/core/src/main/java/org/teavm/backend/wasm/intrinsics/reflection/ReflectionMetadataGenerator.java
+++ b/core/src/main/java/org/teavm/backend/wasm/intrinsics/reflection/ReflectionMetadataGenerator.java
@@ -884,7 +884,8 @@ public class ReflectionMetadataGenerator {
         function.add(thisVar);
         var argsVar = new WasmLocal(objectArrayClass.getType(), "args");
         function.add(argsVar);
-        var argsDataVar = new WasmLocal(dataField.getUnpackedType(), "argsData");
+        var argsDataVar = new WasmLocal(((WasmType.Reference) dataField.getUnpackedType()).asNullable(),
+                "argsData");
         var body = function.getBody().builder();
         var args = new WasmInstructionList().builder();
 
--- a/tests/src/test/java/org/teavm/classlib/java/lang/reflect/MethodTest.java
+++ b/tests/src/test/java/org/teavm/classlib/java/lang/reflect/MethodTest.java
@@ -281,6 +281,14 @@ public class MethodTest {
         assertEquals(25, result);
     }
 
+    @Test
+    @SkipPlatform(TestPlatform.C)
+    public void asyncReflectionWithSeveralParameters() throws Exception {
+        var method = ClassWithAsyncMethodWithSeveralParameters.class.getMethod("foo", int.class, int.class);
+        var result = method.invoke(null, 23, 42);
+        assertEquals(67, result);
+    }
+
     private void callMethods() {
         new Foo().bar(null);
         new Foo().baz();
@@ -529,6 +537,17 @@ public class MethodTest {
             Thread.sleep(1);
             ++x;
             return x;
+        }
+    }
+
+    static class ClassWithAsyncMethodWithSeveralParameters {
+        @Reflectable
+        public static int foo(int x, int y) throws InterruptedException {
+            Thread.sleep(1);
+            ++x;
+            Thread.sleep(1);
+            ++y;
+            return x + y;
         }
     }
 }


konsoletyper:
I believe it's not an optimal way to solve the issue you are trying to solve here. Can you provide an end-to-end example with the reproduction, without the synthetic wasm generator? I believe, there was the code that disabled generation of non-null variables for suspended function. If it's not the case for some reason, it's better to fix the root reason rather than trying to patch coroutine transformation.

lucas-viva:
Thanks @konsoletyper!

Here is an end-to-end reproduction without the synthetic generator:
https://github.com/lucas-viva/gdx-teavm-1246-repro

It is an ordinary libGDX game built with the gdx-teavm backend: a skin-based UI, a background
loading thread, and a button listener that `join()`s the loader. Compiled against current
master it is rejected at load time:

```
CompileError: WebAssembly.compileStreaming(): Compiling function #2994:
"com.badlogic.gdx.scenes.scene2d.ui.SplitPane::<..." failed:
uninitialized non-defaultable local: 2 @+845479
```

Claude pointed me two possible places to fix it:

1. In the transformation (this PR): map non-defaultable locals to nullable ones and re-narrow
   each original read with a cast. This covers every producer - the classlib-reflection caller
   generator today, and any other body intrinsic that declares a non-null local and ends up
   suspendable.
2. In `ReflectionMetadataGenerator.generateCaller` alone: stop caching the unpacked array data
   in a non-null local (read it inline, as the introspection callers do). Narrower, but the
   transformation remains unable to handle non-null locals from any other generated body.

Claude recommended 1 because the transformation is the single place that sees every function it is
about to rewrite. The test uses a small generator because that is the only in-tree way to feed
such a function to the transformation. Happy to rework it as another solution, if you prefer.

Also please note the example will hit another issue with Fiber code. A follow-up fix for that is prepared separately and I will create a PR for it after this PR's resolution.

konsoletyper:
> It is an ordinary libGDX game

Thanks for the feedback, but I believe producing async code is not what you want for game development. I'd got rid of any code that triggers coroutine transformation

konsoletyper:
So this snippet is a thing you need to avoid:

```java
        loader = new Thread(() -> {
            try {
                Thread.sleep(2000);
            } catch (InterruptedException e) {
                Thread.currentThread().interrupt();
            }
        });
        loader.start();
```

I don't know how libGDX works, so I'd consult with the maintainers. Perhaps, there's a better alternative to directly spawning a thread.

lucas-viva:
Thanks for the comments!

I reworked the PR the way you suggested: fix the root reason and leave the coroutine
transformation alone.

You remembered correctly: `WasmGCMethodGenerator` does disable non-nullable locals when a
method is suspendable, and that guard pointed me straight at the real problem. The
reflection callers are not regular method bodies: `ReflectionMetadataGenerator
.generateCaller` builds its wasm body directly and applies the coroutine transformation
itself, so the guard never sees them. When a member has more than one parameter, the caller
cached the argument array's data field in a non-null local - that is the local the validator
rejects. The PR now simply deletes that local and reads the array through the (always
initialized) `args` parameter at each use. The transformation is untouched.

The synthetic generator is gone from the test as well. A plain test turned out to reproduce
it: a `@Reflectable` method with two parameters that calls `Thread.sleep`, invoked via
`Method.invoke`. On current master it fails with:

```
CompileError: WebAssembly.compileStreaming(): Compiling function #153:
"org.teavm.vm.WasmAsyncTest$ReflectiveTarget::co..." failed:
uninitialized non-defaultable local: 2
```

Good point about the thread, but unfortunately avoiding it does not help: the thread is not
what triggers this. Any blocking primitive reachable from a method of a reflectable class
does. A bare `Thread.sleep(...)` in the button listener reproduces the same failure with no
thread anywhere. And since gdx-teavm registers all of
`com.badlogic.gdx.scenes.scene2d.**` for reflection by default (that is how `Skin`/JSON UI
loading works), it is hard for an application to guarantee the combination never occurs. You
are right that keeping the async surface small is good advice for performance, and we follow
it where we can - this fix is for the cases that slip through, where today the result is a
validation error pointing at a generated wrapper the user has never seen.

For context on where this series comes from: we are porting Kakele Online
([kakele.io](https://kakele.io)), a live libGDX MMORPG, to the wasm-gc backend - and it has
been worth it: the same code that runs on the JS backend at 33~ fps runs locked at 60 fps as wasm-gc
on old phones. The remaining fixes we needed are prepared as separate branches, we will submit a few more PRs
one at a time for you to review: https://github.com/lucas-viva/teavm. We would much rather
converge on upstream than keep carrying a fork of our own.

Happy to adjust the fix or the test if you would like a different shape.

konsoletyper:
> A bare Thread.sleep(...) in the button listener reproduces

Sure, that's why you should avoid this as well. I believe, libGDX relies on some sort of event queue, so all delays should be performed by using libGDX-specific APIs rather than on Thread.sleep.

> And since gdx-teavm registers all of com.badlogic.gdx.scenes.scene2d.**

Then it's a problem of gdx-teavm, which should be addressed there.

lucas-viva:
Following up on the gdx-teavm point: you were right that it belongs there as well. I measured what the default registration costs and filed [xpenatan/gdx-teavm#193](https://github.com/xpenatan/gdx-teavm/issues/193) - the defaults expose every member of 249 classes (3,204 methods and constructors), while `Skin`/`Json` and `Pools` only ever use fields and no-arg constructors. Narrowing our own build to that surface cut the wasm module by 18% and more than halved the TeaVM build step, with no behaviour change.

It also removes the 693 members with two or more parameters, which are exactly the wrappers this PR is about, so with the narrowed surface our game no longer hits the validation error at all. This PR is therefore not blocking us any more. I will leave it to you whether that case is worth fixing (please check the code again, I removed the changes in the coroutine transformation class) - merge or close, either is fine by me.

For what it is worth, I also verified the reworked fix against the full game with the stock gdx-teavm reflection surface, not just the small reproduction: it validates and runs.

Thanks for the reviews. The remaining changes we need are unrelated to this one and I will submit them one at a time.

konsoletyper:
I see. I don't have idea why this reflection description mechanism exists in gdx-teavm in the first place. TeaVM already supports very convenient compiler extension mechanist exactly for the same purpose, and it's much more flexible: https://teavm.org/docs/runtime/compiler-extensions.html (`ReflectionPolicy` is perhaps one that you need).

tests/src/test/java/org/teavm/vm/WasmAsyncTest.java
```diff
@@ -66,6 +67,16 @@ public void suspendWithValueTeedIntoWiderLocal() {
         assertEquals(54, teeThenSuspend());
     }
 
+    @Test
```
konsoletyper: I believe, `MethodTest` has something similar. I would recomment moving it there, since this test is not Wasm-specific (but was reproduced only in Wasm backend)

lucas-viva: Moved, thanks. The new test sits next to `MethodTest.asyncReflection` and invokes a suspendable method with two parameters, like the existing one. I gave it its own nested class so the neighbouring test's module is unaffected. `WasmAsyncTest` is untouched now

core/src/main/java/org/teavm/backend/wasm/intrinsics/reflection/ReflectionMetadataGenerator.java
```diff
@@ -884,7 +884,6 @@ private WasmFunction generateCaller(MethodReader method) {
         function.add(thisVar);
         var argsVar = new WasmLocal(objectArrayClass.getType(), "args");
         function.add(argsVar);
-        var argsDataVar = new WasmLocal(dataField.getUnpackedType(), "argsData");
```
konsoletyper, line 887: It's strange that you modify (deoptimize) the code below, while it would be easier to just turn this local into nullable

lucas-viva: Done. The local is now declared nullable. Thanks!

PR #1251, opened 2026-09-07
Fix Character.toTitleCase mappings and identity

cstainton:
Hi! I came across a small issue in `Character.toTitleCase()` while checking Unicode handling in a project using TeaVM.

The method compares the mapped code point with itself, so it always takes the uppercase fallback. For example, `ǳ` becomes `Ǳ` rather than its title-case form, `ǲ`.

This fix preserves dedicated title-case mappings and characters already in title case, while keeping the usual uppercase fallback for ordinary letters. Invalid code points are returned unchanged, matching the Java API.

I've added regression coverage for both overloads, ordinary letters, supplementary characters and invalid inputs. All 12 Character tests and Checkstyle pass.

Validation used the JavaScript backend; Wasm GC, C and optimized variants were disabled for this focused run:

```sh
./gradlew :tests:test --tests org.teavm.classlib.java.lang.CharacterTest :classlib:checkstyleMain :tests:checkstyleTest -Pteavm.tests.js=true -Pteavm.tests.wasm-gc=false -Pteavm.tests.c=false -Pteavm.tests.optimized=false
```

Thanks for all your work on TeaVM - hoping to contribute a few useful libraries to the TeaVM ecosystem soon.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/lang/TCharacter.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/lang/TCharacter.java
@@ -279,11 +279,12 @@ public class TCharacter extends TObject implements TComparable<TCharacter> {
     private static native StringResource acquireUpperCaseMapping();
 
     public static int toTitleCase(int codePoint) {
-        codePoint = mapChar(getTitleCaseMapping(), codePoint);
-        if (codePoint == codePoint) {
-            codePoint = toUpperCase(codePoint);
+        if (!isValidCodePoint(codePoint) || isTitleCase(codePoint)) {
+            return codePoint;
         }
-        return codePoint;
+        // TeaVM stores only title-case mappings that differ from uppercase.
+        int titleCase = mapChar(getTitleCaseMapping(), codePoint);
+        return titleCase != codePoint ? titleCase : toUpperCase(codePoint);
     }
 
     public static char toTitleCase(char c) {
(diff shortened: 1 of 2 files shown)


classlib/src/main/java/org/teavm/classlib/java/lang/TCharacter.java
```diff
@@ -279,11 +279,12 @@ private static CharMapping getUpperCaseMapping() {
     private static native StringResource acquireUpperCaseMapping();
 
     public static int toTitleCase(int codePoint) {
-        codePoint = mapChar(getTitleCaseMapping(), codePoint);
-        if (codePoint == codePoint) {
-            codePoint = toUpperCase(codePoint);
+        if (!isValidCodePoint(codePoint) || isTitleCase(codePoint)) {
```
konsoletyper, line 282:
```
    public static int toTitleCase(int codePoint) {
        var mappedCodePoint = mapChar(getTitleCaseMapping(), codePoint);
        if (codePoint == mappedCodePoint) {
            codePoint = toUpperCase(codePoint);
        }
        return codePoint;
    }
```

konsoletyper: Sorry, my bad. This does not work

cstainton: Thank you very much - I have another request that's a bit more involved that covers junit rule support (so, an addition), I will send it later. Cheers !

PR #1253, opened 2026-09-08
Support JUnit TestRule in compiled TeaVM tests

cstainton:
# Support JUnit `TestRule` in compiled TeaVM tests

I found that `TestRule` instances declared through `@Rule` fields and methods
were silently ignored in tests compiled by TeaVM. Tests could pass without
running the rule's setup, cleanup, or assertions. This adds support for applying
those rules around the test's setup, body, and teardown.

The change handles rule fields, no-argument rule methods, inherited rules, and
`@Rule(order = ...)`. Rules receive a `Description` naming the test class and
method. Teardown failures are propagated rather than only printed.

It also adds `TConcurrentLinkedQueue`, which JUnit needs when constructing a
`Description`. This uses `TAbstractQueue` and `TArrayDeque`. It is intended for
single-threaded use and does not provide the JVM queue's concurrency or weakly
consistent iterator guarantees.

This focuses on per-test `TestRule` execution, which our downstream tests use.
Supporting `MethodRule` would also require making the test's reflective
`Method` available to its `FrameworkMethod`. Supporting `@ClassRule` would
require applying rules once around the whole class, including class-level setup
and teardown. Those need additional compiler and runner work beyond the
per-test execution added here; they are not inherently unsupported by TeaVM.

The JVM comparison runner is unchanged; the rule tests use `@SkipJVM` because
that runner invokes test methods directly without applying rules.

Downstream, we use a small jar containing the patched classes, placed ahead of
TeaVM's jars on the test classpath. An upstream release containing this support
would let us remove that workaround.

## Tests

All 29 selected tests passed: 12 queue tests, six rule tests, and the 11 existing
`BufferedWriterTest` tests. They cover queue operations, rule execution and
ordering, test descriptions, reverse unwinding, and teardown failure.

Validation used JavaScript and Wasm GC, including optimised builds. The queue
and BufferedWriter tests also ran on the JVM. Main-source and test-source
Checkstyle checks passed.

The C backend, full test suite, and combined test-body/teardown failure case
have not been tested.

Thanks for reviewing this contribution.

### DIFF
--- a/classlib/src/main/java/org/teavm/classlib/java/util/concurrent/TConcurrentLinkedQueue.java
+++ b/classlib/src/main/java/org/teavm/classlib/java/util/concurrent/TConcurrentLinkedQueue.java
@@ -1,0 +1,132 @@
+/*
+ *  Copyright 2026 Carl Stainton.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.teavm.classlib.java.util.concurrent;
+
+import org.teavm.classlib.java.io.TSerializable;
+import org.teavm.classlib.java.lang.TNullPointerException;
+import org.teavm.classlib.java.util.TAbstractQueue;
+import org.teavm.classlib.java.util.TArrayDeque;
+import org.teavm.classlib.java.util.TCollection;
+import org.teavm.classlib.java.util.TIterator;
+import org.teavm.classlib.java.util.TQueue;
+
+/**
+ * A first-in, first-out queue for code compiled by TeaVM in a single-threaded runtime.
+ *
+ * <p>This class supplies TeaVM's implementation of {@link java.util.concurrent.ConcurrentLinkedQueue}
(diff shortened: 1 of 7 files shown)


konsoletyper:
That's definitely no.

1. This PR suggest major changes in JUnit and in classlib emulation, which I believe should not be done. I would usually suggest to split this
2. The naive assumption that you can just implement multi-threaded collections as single-threaded, because "JavaScript is a single-threaded environment" is wrong. TeaVM emulates threads with its own coroutines, so such code should be written carefully. Take a look at TConcurrentHashMap
3. In case you generated this PR with a coding agent, than it's absolutely no. I can't accept any changes to classlib module, written by an agent. Rationale: LLM behind it could have seen OpenJDK code during training, which makes it possible that the model writes code that uses patterns and approached, taken by OpenJDK, *which is a licensing violation risk*.
4. I also believe that with the proper approach you can move further without emulation of extra multithreaded collections.

konsoletyper:
I believe, `Description` class can be patched to replace any occurrence of concurrent collections with simple ones. A `ClassHoldeTransformer` can be used for this purposes. You need both to replace class in `ConstructInstruction` as well as method's owning class in `InvokeInstruction` (I believe, applying it to `<init>` methods should be enough).

cstainton:
Hi,

Thanks for the tip - sorry I had my eye on the ball of supporting JUnit Rules and wrote the naive just-enough substitute for the JDK class. My notes from the time cover thoughts and questions about classlib contributions and SupplierPolicy use so yes the suitability for wider consumption was considered but not deeply enough.

Instead (of this PR) I've implemented it downstream as an additional class path library that patches Description and TestEntryPoint and must be placed before TeaVM's JUnit dependency. I've written a suspension safe Collection which doesn't need as much of the api of ConcurrentLinkedQueue - that comes in via the patched Description. 

Armed with the patching approach I was motivated to provide MethodRule support too, which involved delving into the Reflection support, and I encountered a small issue there which I'll raise as a separate PR.

Thanks

PR #1256, opened 2026-09-14
wasm gc: nullable types for the labels the coroutine transformation creates

lucas-viva:
A value with a non-null reference type on the stack beneath a conditional whose arm suspends
makes the generated module fail validation:

```
CompileError: WebAssembly.compileStreaming():
type error in fallthru[1] (expected (ref 9), got (ref null 9))
```

The restore blocks built by the coroutine transformation map their signatures through
`mapToNullableTypes`, because restored and default values are nullable. The labels created for
optimized blocks and flattened conditionals used the recorded stack types directly, so the
restore path delivers a nullable value to a non-null label. These signatures now go through the
same `mapToNullableTypes`, and the recorded stack is reset to the declared output types so later
suspension points snapshot what validation sees. A non-null value always satisfies a nullable
label, so existing producers are unaffected.

The test in `AsyncTest` holds a string constant beneath a suspending conditional; it fails on
current master with the error above.


konsoletyper:
As I already explained, it makes no sense to support non-nullable types in coroutines methods. My guess was that nullability information could hint a Wasm optimizer to produce a better code. It's a quite questionable assumption, but for coroutine methods it makes no sense: suspension points add so much overhead that it simply makes no sense to keep nullability information. I stand strictly against introduction of extra logic to CoroutineTransformer. I already explained it in previous reviews.

I know that you are likely using AI agent and just paste my comments directly to it. But please, @lucas-viva, PLEASE! I address you personally: just read my comments and whatever code the LLM produces. Don't be an interlayer between your coding agent and me. I have a coding agent myself, if you want to help project, donate so that I could by more tokens and review such PRs without need of your time and efforts. Otherwise, I want you to bring some value to PRs you post.

lucas-viva:
Hi @konsoletyper, sorry, my bad.

I forgot you mentioned we shouldn't modify the coroutine transformer.

Yes, I'm using Claude to guide me through the modifications, and I will be extra careful moving forward, so we don't waste your time reviewing bad solutions.

Moving forward, I will change our strategy: I will first open an issue describing the problem, add a test and/or a small reproducible example and wait for your feedback on it. Then, I will wait you to write the solution or I can do the PR for you to review too, whichever you prefer.

For this PR, do you want me to close it or leave it open with the test code only?

By the way, we are already sponsoring the project (I'm part of ViVa Games).

Thanks for the feedback.

konsoletyper:
> Moving forward, I will change our strategy: I will first open an issue describing the problem, add a test and/or a small reproducible example and wait for your feedback on it. Then, I will wait you to write the solution or I can do the PR for you to review too, whichever you prefer.

Thanks! I like this idea. Claude is good at writing reproducers. Just ask it to write a test and make sure it fails. You can leave this PR, I'll take a test from it and try to fix the issue myself. Alternatively, you can write PR and add a notice that it's ok to modify this PR from my side rather than to review it.

konsoletyper:
I also want to again raise concern here. With game engine you don't need to write any async code. It's good for the project that you encounter and report issues like this, but it's not good for you at all. What you expect is a good performance, which is never the case with TeaVM coroutines.

lucas-viva:
> I also want to again raise concern here. With game engine you don't need to write any async code. It's good for the project that you encounter and report issues like this, but it's not good for you at all. What you expect is a good performance, which is never the case with TeaVM coroutines.

Thank you for the concern. We do try to minimize the async code. Our game is multi-platform (PC, Android, iOS), so we have a core module that adopts a blocking model that is shared with each target and can't practically be rewritten for one backend. Note also that the current wasm build runs on 60 fps on mid-end phones, where JS gives ~30, so the overhead isn't noticeably hurting us.

konsoletyper:
> so we have a core module that adopts a blocking model that is shared with each target and can't practically be rewritten for one backend

It's possible. You have a separate entry point per backend. So what you need is to hide all platform-specific code behind an interface, then implement this interface for JVM and TeaVM separately and pass corresponding implementation to the game logic in corresponding entry point.

konsoletyper:
> where JS gives ~30

Instead of measuring FPS, profile and identify what's causing this. Perhaps, it's reliance on `long` somewhere in critical path.

Also note that when you start using async code, it does not affect performance locally. The worst thing about it is that it infects *entire* application with async, which not always easy to predict, so performance can be fragile - one innocent change can reduce performance dramatically.

konsoletyper:
Fixed

lucas-viva:
> > where JS gives ~30
> 
> Instead of measuring FPS, profile and identify what's causing this. Perhaps, it's reliance on `long` somewhere in critical path.
> 
> Also note that when you start using async code, it does not affect performance locally. The worst thing about it is that it infects _entire_ application with async, which not always easy to predict, so performance can be fragile - one innocent change can reduce performance dramatically.

Thank you again for your feedback and guidance and pushing back on async code.

I went back and checked how much of our game was being transformed into suspension points due to async code. It was around 26k methods out of 29k, about 90%. Not because we use threads: one blocking InputStream.read override (a WebSocket-backed socket in our browser client) made java.io.InputStream.read an async family, so every unrelated stream reader (like BufferedReader, protobuf's CodedInputStream, GZIPInputStream, our asset loading, etc) became a suspension point, and every caller of those a state machine. It came to a point where even IntArray.get(int) was being compiled as a coroutine.

Moving that blocking read behind an interface only our packet reader calls brought it down significantly. Now we have 69 async methods, all confined to our connection code. With that small surface, none of the three WASM-GC coroutine bugs we had been hitting reproduce any more in our game - this PR included. Our game now builds and runs on master on both targets (logged in and played) with only an unrelated classlib fix on top (I will open an issue for that).

Both JS and WASM-GC targets run on 60 fps now.

I will still open issues with the reproducers for the three WASM-GC coroutine bugs for you to take a look too.

Thanks for the pushback, it pointed at a much better state of the code.

